PR #15, opened 2020-07-22
[MODORDERS-406] Prevent the creation of an encumbrance for an inactiv…

KreiDer:


KreiDer:
Verified on snapshot
image

PR #16, opened 2020-07-23
[MODINVOICE-174] - API tests

AndreiMPV:


KreiDer:
Verified on snapshot
image

PR #28, opened 2020-08-06
MODFIN-138 Update /finance/budgets POST and PUT APIs

pautau-s:
- update budget-expense-classes.feature to test updated /finance/budgets API

Verified on folio-testing:
image


AndreiMPV:
Tested on snapshot
image

PR #48, opened 2020-08-28
MODFIN-147 Budget allocation calculation out of sync with transaction log

pautau-s:
https://issues.folio.org/browse/MODFIN-147
Verified on folio-testing:
image


AndreiMPV:
image

PR #51, opened 2020-09-10
[MODINVOICE-188] Update tests with adjustments

KreiDer:
image
Verified on testing


pautau-s (approved):
Verified on folio-testing:
image

PR #52, opened 2020-09-17
[MODFISTO-142] Restrict transfer money if budget doesn't have enough …

KreiDer:
image
Verified on testing


pautau-s (approved):
Verified on folio-testing
image

PR #53, opened 2020-09-17
MODINVOICE-181 Update Pending payments when exchange rate is edited

pautau-s:
https://issues.folio.org/browse/MODINVOICE-181

Local run:
image


AndreiMPV:
image

PR #54, opened 2020-09-21
MODFIN-153 added transaction-summaries-crud.feature

grabsefx:
### Purpose
https://issues.folio.org/browse/MODFIN-153

### Approach
Verified on folio testing
image


pautau-s (approved):
Verified on folio-testing
image

PR #56, opened 2020-09-23
Add test on resumptionToken for marc21_withholdings metadata prefix 

IlliaDaliek:


mrdpopov:
If this is a copypaste from an existing test, please consolidate it in one test and use templates:  <metadataPrefix>

IlliaDaliek:
> If this is a copypaste from an existing test, please consolidate it in one test and use templates:

Okay, will do it a bit later and then merge.

PR #59, opened 2020-09-28
[MODINVOICE-148] Add integration test for order invoice relation

KreiDer:
Signed-off-by: Yauheni Kuzmianok <yauheni_kuzmianok@epam.com>
image


pautau-s (approved):
Verified on folio-testing
image

PR #68, opened 2020-10-29
MODORDERS-437 added expense classes checks

grabsefx:
https://issues.folio.org/browse/MODORDERS-437

image


pautau-s (approved):
Verified on folio-snapshot

PR #76, opened 2020-11-29
MODINVOICE-214 - Support ability display lock and calculated totals

AndreiMPV:


AndreiMPV:
image

pautau-s (approved):
Verified on folio-testing:
image

PR #84, opened 2021-01-06
MODFIN-168 Update logic with financial detail in the Ledger summary to improve the users ability to manage financial activity

pautau-s:
https://issues.folio.org/browse/MODFIN-168

image


AndreiMPV:
image

PR #98, opened 2021-02-01
MODINVOICE-222 adjust remaining amount checks

grabsefx:
## Purpose
* https://issues.folio.org/browse/MODINVOICE-222

## Approach

adjust remaining amount checks
image


pautau-s (approved):
image

PR #100, opened 2021-02-01
MODORDERS-443 Update logic for supporting manually exchange rate for …

KreiDer:
[MODORDERS-443](https://issues.folio.org/browse/MODORDERS-443)
Update logic for supporting manually exchange rate for purchase order line


KreiDer:
image

PR #103, opened 2021-02-04
[MODFISTO-206] Test allowable encumbrance and expenditure restrictions

damien-git:
[MODFISTO-206](https://issues.folio.org/browse/MODFISTO-206)
Test allowable encumbrance and expenditure restrictions.

Screenshot_20210204_144213


pautau-s (approved):
image

PR #108, opened 2021-02-11
[MODORDERS-471] Make sure allowable encumbrance restriction is based on total funding

damien-git:
[MODORDERS-471](https://issues.folio.org/browse/MODORDERS-471)

Added tests to `create-order-that-has-not-enough-money.feature` to check that open order will pass if and only if the budget has enough available money based on total funding (including netTransfers).
Screenshot_20210211_111501


AndreiMPV:
image

PR #109, opened 2021-02-12
MODFISTO-222 Budget created for inactive fund

pautau-s:
[MODFISTO-222](https://issues.folio.org/browse/MODFISTO-222)
image


pautau-s:
image

PR #111, opened 2021-02-17
MODFISTO-224 Calculate transaction amount based on fundDistribution percentage value

pautau-s:
https://issues.folio.org/browse/MODFISTO-224
image


AndreiMPV:
image

PR #119, opened 2021-03-23
MODORDERS-482 Cannot split funds on a PO with an odd number of pennies as the cost

pautau-s:
https://issues.folio.org/browse/MODORDERS-482
image


AndreiMPV:
image

PR #131, opened 2021-04-12
MODINVOICE-232 - split odd number of pennies

grabsefx:
## Purpose
https://issues.folio.org/browse/MODINVOICE-232

## Approach
* Created test for odd number of pennies
image


AndreiMPV:
image

PR #147, opened 2021-05-14
Fat-141: Implement API Karate tests for mod-data-export - PART 2

IlliaDaliek:
Jira: https://issues.folio.org/browse/FAT-141

### Purpose
Add tests related to creating file definitions, uploading files and exporting the records.


IlliaDaliek:
It is not all of the tests, I will push another commit with the second part soon.

mod-data-export/src/test/resources/domain/dataexport/features/file-definition.feature
```diff
@@ -0,0 +1,64 @@
...
+    * url baseUrl
+
+    * callonce login testAdmin
+    * def okapiAdminToken = okapitoken
+
+    * callonce login testUser
+    * def okapiUserToken = okapitoken
+
+    # load variables
+    * callonce variables
```
bohdan-suprun, line 13: Does it make sense to make the step name more descriptive and remove the comment? e.g. `load callonce variables`

mod-data-export/src/test/resources/domain/dataexport/features/file-definition.feature
```diff
@@ -0,0 +1,64 @@
...
+    And configure headers = headersUserOctetStream
+    And request karate.readAsString('classpath:samples/file-definition/test-export.<uploadFormat>')
+    When method POST
+    Then status 200
+    And match response.jobExecutionId == '#present'
+    And match response.status == 'COMPLETED'
+    And match response.uploadFormat == '<uploadFormat>'
+
+    Examples:
+      | uploadFormat | id                  |
```
bohdan-suprun, line 61:
```suggestion
      | fileDefinitionId | uploadFormat|
```

I'd change it to be in the same order and have the same name as other tests.

mod-data-export/src/test/resources/domain/dataexport/features/file-definition.feature
```diff
@@ -0,0 +1,64 @@
...
+      | csv          | csvFileDefinitionId |
+      | cql          | cqlFileDefinitionId |
```
bohdan-suprun, line 63: Do you know if it is possible to specify these examples, globally for all test cases just to reduce duplication?

mod-data-export/src/test/resources/domain/dataexport/features/mapping-profiles.feature
```diff
@@ -83,3 +83,51 @@ Feature: Test mapping profiles
     Then status 403
     And match response contains 'Deletion of default mapping profile is forbidden'
 
+  Scenario: should return UnprocessableEntity response when post mapping profile with invalid transformations - invalid tag
+    Given path 'data-export/mapping-profiles'
+    * print read('classpath:samples/mapping-profile/mp-transformation_invalid_tag.json')
+    And def invalidMappingProfile = read('classpath:samples/mapping-profile/mp-transformation_invalid_tag.json')
+    And request invalidMappingProfile
+    When method POST
+    Then status 422
```
bohdan-suprun, line 92: Could you add an assertion that 422 was due to `invalid tag` but not due to an another reason? 

If yes, please update all similar tests.

mod-data-export/src/test/resources/domain/dataexport/features/mapping-profiles.feature
```diff
@@ -83,3 +83,51 @@ Feature: Test mapping profiles
...
+
+  Scenario: should return OK response when post mapping profile with empty transformation and holdings as transformation record type
+    Given path 'data-export/mapping-profiles'
+    And def validMappingProfile = read('classpath:samples/mapping-profile/mp-empty-transformation-holding-record-type.json')
+    And set validMappingProfile.id = uuid()
+    And set validMappingProfile.name = randomString(10)
+    * print validMappingProfile
+    And request validMappingProfile
+    When method POST
+    Then status 201
```
bohdan-suprun, line 123: Should we then `GET` the entity after and verify it's state?

mod-data-export/src/test/resources/domain/dataexport/features/file-definition.feature
IlliaDaliek: Thank you, but I did a common table for "scenario outline", therefore this change is not required anymore. The same relates to the comment below.

mod-data-export/src/test/resources/domain/dataexport/features/file-definition.feature
IlliaDaliek: See comment above.

mod-data-export/src/test/resources/domain/dataexport/features/mapping-profiles.feature
IlliaDaliek: There a simple regex that doesn't allow to pass some invalid values and the error message is common for all cases, so I cannot add a check that the invalid tag led to the error.
error exmple
reason

mod-data-export/src/test/resources/domain/dataexport/features/file-definition.feature
IlliaDaliek: Function renamed to "loadTestVariables"

mod-data-export/src/test/resources/domain/dataexport/features/mapping-profiles.feature
IlliaDaliek: The POST method here returns the created entity as JSON and we can verify it here without additional GET call, I added some cheks. As well, the suggested testing behavior with follow-up GET calls already present above in a file. I just added specific cases.

mod-data-export/src/test/resources/domain/dataexport/features/export.feature
```diff
@@ -0,0 +1,144 @@
...
+    And configure headers = headersUserOctetStream
+    And request karate.readAsString('classpath:samples/file-definition/<fileName>')
+    When method POST
+    Then status 200
+    And match response.jobExecutionId == '#present'
+    And match response.status == 'COMPLETED'
+    And match response.uploadFormat == '<uploadFormat>'
+    And match response.sourcePath == '#present'
+    And def jobExecutionId = response.jobExecutionId
+    And call pause 1000
```
bohdan-suprun, line 48: Could you please add a comment or name the step properly to understand why do we need the pause?

Is it pause of the test or some module pause?

mod-data-export/src/test/resources/domain/dataexport/features/export.feature
```diff
@@ -0,0 +1,144 @@
...
+
+  Scenario: Should return transformation fields
+    Given path 'data-export/transformation-fields'
+    When method GET
+    Then status 200
+    And assert response.transformationFields.length > 0
+
+  Scenario: should clean expired jobs if exist and return 204
+    Given path 'data-export/expire-jobs'
+    When method GET
```
bohdan-suprun, line 101: Expired jobs here refers to the ones created by `Scenario Outline: test upload file and export flow.`, right?

If yes, we should make sure that the order is preserved, but my personal preference would be to extend this test so that it first creates some jobs and than verifies that they removed.

It is up to you to decide.

mod-data-export/src/test/resources/domain/dataexport/features/mapping-profiles.feature
bohdan-suprun: Can you verify `value` at least. I assume it has different value depending on the case?

mod-data-export/src/test/resources/domain/dataexport/features/export.feature
IlliaDaliek: The problem here that when we upload a file it returns 204 immediately and starts the processing file in the background and the next API call requires this file to be uploaded but it may need some time to be done. That is why we need to wait a bit, but a better way is to use the 'retry until' condition by which we can check that the related to uploading file job has status 'completed', if it doesn't then we wait 500 ms and we make a call again to be sure that the job has completed status before invoking further calls that depend from the status of this job. So, I plugged in the "retry until" statement. As well I updated the description regarding these delays.

mod-data-export/src/test/resources/domain/dataexport/features/export.feature
IlliaDaliek: Actually, this scenario doesn't relate to the export flow and the test defined separately from the main flow in "scenario outline". I removed the scenario cause it requires a lot of time to make the job expired. This functionality covered by rest assured tests inside the module and the logic of cleaning jobs/files doesn't contain dependencies from external calls, therefore it can be dropped from integration tests I think. Your thoughts?

bohdan-suprun (approved):
Looks good!

PR #156, opened 2021-05-24
FAT-211: Implement base test for mod-search

bohdan-suprun:
* Add a data set up feature;
* Add a data destroy feature;
* implement basic search by title.


pfilippovEPAM (approved):
Looks good to me.

mod-search/src/main/resources/karate-config.js
```diff
@@ -0,0 +1,72 @@
...
+
+  var env = karate.env;
+
+  // specify runId property for tenant postfix to avoid close connection issues
+  // once we run tests again
+  var runId = karate.properties['runId'];
+
+  var config = {
+    baseUrl: 'https://falcon-okapi.ci.folio.org',
+    admin: {tenant: 'diku', name: 'diku_admin', password: 'admin'},
```
hjiebsco, line 13: Hardcoded password?

hjiebsco (commented):
Is there a way to share `karate-config.js` with other modules? Also, should diku password be embedded?

mod-search/src/main/resources/karate-config.js
bohdan-suprun: Yes, all modules has hardcoded passwords. No way to inject variables exists now.

bohdan-suprun: > Is there a way to share karate-config.js with other modules?

The current approach is that each module is isolated and defines it's own `karate-config.js`

hjiebsco (commented):
I am not a fan of hardcoding password and checking into to GitHub. I will defer this to others.

bohdan-suprun:
@hjiebsco, I have created https://issues.folio.org/browse/FOLIO-3178 - to adjust jenkins job and https://issues.folio.org/browse/FAT-916 to apply the properties after the job is updated. 

I'm going to merge this PR.

PR #182, opened 2021-06-11
FAT-168 Create test plan for mod-login-saml

steveellis:
Resolves [FAT-168](https://issues.folio.org/browse/FAT-168) which is only to implement the test plan (adding undefined tests to output to the Karate report). You may notice a few TODO comments.  I left those in because I wanted to capture the knowledge that they contained. Resolving them (and I suspect others that will be similar) can be handled in the next phase of Karate test integration (actually implementing all the tests) and probably will require some discussion of the best approach for resolution with someone who knows the SAML code well. That work is covered in [FAT-169](https://issues.folio.org/browse/FAT-169).


mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature
```diff
@@ -3,11 +3,143 @@ Feature: Login SAML
...
+    # (this is karate's 'schema validation') and the value.
+    Then match response ==
+    """
+    {
+      active: '#boolean'
+    }
+    """
+    Then match response.active == false
+
+  # TODO In the rest assured tests this is 400 active true. But here it is active false 200. Which is correct?
```
hjiebsco, line 24: I do not see rest assured tests returns `400 active true`. Are you referring to this [code](https://github.com/folio-org/mod-login-saml/blob/v2.2.0/src/test/java/org/folio/rest/impl/SamlAPITest.java#L133-L164)? It returns 400 only when `x-okapi-tenant` header is missing.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSamlUnauthenticated.feature
```diff
@@ -0,0 +1,17 @@
...
+  Background:
+    * url baseUrl
+    * configure headers = { 'x-okapi-tenant': #(testTenant) }
+
+  @Undefined
+  Scenario: Check endpoint bad request
+  * print 'undefined'
+# TODO In the rest assured tests, this is 400, not 200 active false. I believe this is because when the tests
+# are run against vertx directly the request without the headers responds 400 (bad request) -- never reaching
+# getSamlCheck. But when it is run through okapi the request is routed to the getSamlCheck method.
```
hjiebsco, line 14: Yes and no. RMB generated API implementation checks for `x-okapi-tenant` header. If updating the rest assured test with `x-okapi-tenant` header, it will return 200 as well.

hjiebsco (approved):

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature
steveellis: Yes that's right. That's a typo in my comment. It should read ```200 active true```. If I run the test that is [commented out](https://github.com/folio-org/folio-integration-tests/blob/28fbcf919a1bf87cdbed0bc1c56dd257f804af5e/mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature#L28) I get this result, which I think I should instead be ```active: true``` to match the rest assured test.

```
17:28:16.260 request:
1 > GET https://folio-testing-okapi.dev.folio.org:443/saml/check
1 > Accept: application/json, text/plain
1 > Accept-Encoding: gzip,deflate
1 > Connection: Keep-Alive
1 > Content-Type: application/json
1 > Host: folio-testing-okapi.dev.folio.org:443
1 > User-Agent: Apache-HttpClient/4.5.12 (Java/11.0.10)
1 > x-okapi-tenant: testenant6106266363388642304
1 > x-okapi-token: eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJ0ZXN0LXVzZXIiLCJ1c2VyX2lkIjoiMDAwMDAwMDAtMTExMS01NTU1LTk5OTktOTk5OTk5OTk5OTkyIiwiaWF0IjoxNjIzNDQ2ODk2LCJ0ZW5hbnQiOiJ0ZXN0ZW5hbnQ2MTA2MjY2MzYzMzg4NjQyMzA0In0.IhvX1lVXsdrLXjvHdvDEmF03M3ujhThkCwriegZV9Yc
1 > x-okapi-url: http://localhost:9130

17:28:16.398 response time in milliseconds: 137.23
1 < 200
1 < Connection: keep-alive
1 < Content-Type: application/json
1 < Date: Fri, 11 Jun 2021 21:28:16 GMT
1 < Transfer-Encoding: chunked
1 < X-Okapi-Trace: GET mod-login-saml-2.2.1-SNAPSHOT.70 http://10.36.1.135:9146/saml/check : 200 14735us
1 < vary: origin
{
  "active": false
}
```
Maybe this is because there is no IdP configured. I think it's this code [here](https://github.com/folio-org/mod-login-saml/blob/d8e202e0a307d12b2080dd5b76a7afd59463c8d9/src/main/java/org/folio/rest/impl/SamlAPI.java#L556) and probably [this](https://github.com/folio-org/mod-login-saml/blob/d8e202e0a307d12b2080dd5b76a7afd59463c8d9/src/main/java/org/folio/config/SamlClientLoader.java#L64) that are causing ```active: false```.

Part of the problem with these karate tests is that -- at least I haven't figured it out yet -- is they're not that easy to debug. So I'm not actually able to see if [this line](https://github.com/folio-org/mod-login-saml/blob/d8e202e0a307d12b2080dd5b76a7afd59463c8d9/src/main/java/org/folio/config/SamlClientLoader.java#L64)  is being hit by placing a breakpoint or even console logging because the module is being pulled in through maven. Probably there's a way to have maven point a local mod-login-saml project. But anyway you see what I mean...

Let me know what you think. It would be nice to figure out why this test isn't working.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSamlUnauthenticated.feature
steveellis: Actually the ```x-okapi-tenant``` is there in the new karate test (specified in the ````background``` part). This is what the request response looks like:

```
18:28:00.641 request:
1 > GET https://folio-testing-okapi.dev.folio.org:443/saml/check
1 > Accept-Encoding: gzip,deflate
1 > Connection: Keep-Alive
1 > Host: folio-testing-okapi.dev.folio.org:443
1 > User-Agent: Apache-HttpClient/4.5.12 (Java/11.0.10)
1 > x-okapi-tenant: testenant48810419905537024

18:28:00.785 response time in milliseconds: 143.34
1 < 200
1 < Connection: keep-alive
1 < Content-Type: application/json
1 < Date: Fri, 11 Jun 2021 22:28:01 GMT
1 < Transfer-Encoding: chunked
1 < X-Okapi-Trace: [GET mod-authtoken-2.9.0-SNAPSHOT.91 http://10.36.1.135:9134/saml/check : 202 544us, GET mod-login-saml-2.2.1-SNAPSHOT.70 http://10.36.1.135:9146/saml/check : 200 15718us]
1 < vary: origin
{
  "active": false
}
```
I think what is happening is that the old rest assured test doesn't include the  ```x-okapi-header``` and returns 400 bad request because it's missing the header (there's no method/route for it to hit) (that's [what the test wants](https://github.com/folio-org/mod-login-saml/blob/d8e202e0a307d12b2080dd5b76a7afd59463c8d9/src/test/java/org/folio/rest/impl/SamlAPITest.java#L136)). But since we're now going through okapi, we need that header otherwise we'll get 404. Not sure if this test for 400 is really that useful since I think it's only testing the vertx endpoint. I'd argue it could be removed.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature
hjiebsco: The result of `200 active false` is as expected because https://folio-testing.dev.folio.org/ does not have IdP configured.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSamlUnauthenticated.feature
hjiebsco: Agreed. The old rest assured tests are not integration tests. It is not necessary/valid to translate all old rest assured tests to Karate integration tests.

steveellis: This has got me thinking that maybe we should hold off merging this until we've had a chance to discuss more what we're really trying to accomplish with the karate tests. There might be a bigger issue here than this particular test.

My assumption has been that we're wanting to fully replace the rest assured tests with karate tests. So my "test plan" (the ```@Undefined``` tests in this commit) are mostly a 1 for 1 match between the tests in SamlAPITests.java, without really fully understanding why. So someone coming to work on the next phase (implementing the tests) would also assume that is the goal -- which could lead to a lot of unnecessary work.

Looking at the other two modules which have created their test plans (the other [closed tickets here](https://wiki.folio.org/pages/viewpage.action?spaceKey=FOLIJET&title=Karate+API+Integration+Testing)), it would appear that they are operating under the same assumption -- that it should be a 1 for 1 replacement of rest assured with karate. See [this](https://github.com/folio-org/folio-integration-tests/blob/master/mod-patron-blocks/src/main/resources/domain/mod-patron-blocks/features/automatedPatronBlocks.feature) and [this](https://github.com/folio-org/mod-patron-blocks/blob/master/src/test/java/org/folio/rest/impl/AutomatedPatronBlocksAPITest.java) for what I mean. We're just replicating the tests between the two systems, which may be of questionable benefit.

I think for example instead of completely replacing the rest assured tests with karate, an argument could be made to just consider karate "another tool in our toolbelt" -- making it easier to do cross-module integration tests when they are needed, or UI tests, or whatever karate does well. If that's the case, these test plans that we're making would look  very different, and also the amount of work required to implement karate tests would be quite different too.

Maybe this discussion has already taken place and me and the other folks working on the test plans just aren't aware of the background, but also maybe we're rushing into something that isn't that great of an idea.

hjiebsco: Found this [wiki page](https://wiki.folio.org/display/DD/Karate+API+Integration+Tests+Implementation+Guidelines) that might be helpful. It mentions the scope of Karate integration testing
```
Scope 

1. - Identify most important business flows
2. - Add test case to cover a bug condition
3. - Don't write tests for all possible conditions 

```

steveellis: Thanks Hongwei for the background material. It does seem like maybe there's some recognition in this that replicating all the http tests would be overkill. But then again it seems like people are going down that road. We should discuss more maybe in our call tomorrow.

If we agree that doing a 1 for 1 replacement is a bad idea there might be an opportunity to clarify what would constitute a "good use of a karate integration test". Maybe we just need a test or two that captures the whole end to end chain (tenant creation/user create/permissions/login etc) -- which is what these tests are all about.

julianladisch: The goal of testing is to find as many bugs as possible.
Converting existing tests into Karate tests is unlikely to find more bugs.
During bugfest many librarians manually test workflows - these tests might be automated.

There are different levels of testing: https://en.wikipedia.org/wiki/Software_testing#Testing_levels
We have some integration tests at the module level and maven has an integration-test phase for this purpose. These integration tests should stay at the module level for earliest feedback.
Karate is for system testing, and for integration testing in areas no covered by module based integration tests.

julianladisch (approved):

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSamlUnauthenticated.feature
steveellis: Thanks Julien, Hongwei and Adam for your thoughts on this and for the time this morning to discuss. I think we're agreed that we don't want to replace all the rest assured tests with karate but rather want the benefit of both frameworks. I revised the test plan to cover what I think is the critical part of saml to test (a handful of login routes). Let me know if you feel there are others that could benefit from karate but this might be a good start.

julianladisch (approved):

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature
```diff
@@ -3,11 +3,55 @@ Feature: Login SAML
...
-
-  Scenario: SAML Check
-     Given path 'saml/check'
-     And header x-okapi-tenant = testTenant
-     And header x-okapi-token = '#(okapitoken)'
-     When method GET
-     Then status 200
+    * configure headers = { 'Content-Type': 'application/json', 'x-okapi-tenant': #(testTenant), 'x-okapi-token': '#(okapitoken)', 'Accept': 'application/json, text/plain' }
+
+  Scenario: Check endpoint missing okapi url header active is false
```
hjiebsco, line 8: Do not think we need to mention `missing okapi url header`. For Karate integration test, the call goes to Okapi directly, so Okapi url header is not needed. This test scenario is essentially `Check endpoint returns active is false when IdP is not configured`

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature
```diff
@@ -3,11 +3,55 @@ Feature: Login SAML
...
+    {
+      active: '#boolean'
+    }
+    """
+    Then match response.active == false
+
+  # TODO In the rest assured tests this is 200 active true. But here it is active false 200. This is because
+  # the IdP isn't defined for the karate tests. We will need to configure IdP for the karate tests for this to work.
+  @Undefined
+  Scenario: Check endpoint has okapi url header active is true
```
hjiebsco, line 27: Similar to the first test case, this one would be `Check endpoint returns active is true when IdP is configured`

hjiebsco (approved):
Left a comment about `x-okapi-url` header. I assume other tests will cover the most important mod-login-saml workflow.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature
steveellis: Same. Agree this is language is better. Will make the change and merge.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature
steveellis: I agree. I think your wording of it is better. Let me make that change and I'll go ahead and merge.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSaml.feature
steveellis: Hongwei just noticing I didn't reply back to this. It is good to have this confirmed. Thanks for taking a look at it.

PR #198, opened 2021-06-21
FAT-171: Implement mod-user-import API Karate tests based on Test plans 

steveellis:
Resolves issue [FAT-171](https://issues.folio.org/browse/FAT-171). Also, [verified](https://github.com/folio-org/folio-integration-tests/blob/295f478bc8da882060e9386782aff84781571ee4/common/src/main/resources/common/destroy-data.feature#L21) that purge=true is happening in test teardown.


mod-user-import/src/main/resources/domain/mod-user-import/features/userImport.feature
```diff
@@ -19,36 +25,69 @@ Feature: User import
...
+    Given path 'users'
+    And param query = "barcode==" + barcode
 
+    When method GET
+    Then status 200
+    And assert response.users[0].barcode == barcode
+    And assert response.users[0].username == usersToImport.users[0].username
+    And assert response.totalRecords == 1
+    # Assign the user id for use in subsequent scenarios. The id is created when the user is inserted.
+    * def userId = response.users[0].id
```
steveellis, line 70: Apparently sharing variables between scenarios (what I'm doing here) [is not](https://github.com/intuit/karate#script-structure) the "karate way". Want to refactor this so that each scenario stands on its own.

mod-user-import/src/main/resources/domain/mod-user-import/features/userImport.feature
```diff
@@ -19,36 +25,69 @@ Feature: User import
...
+    When method POST
+    Then status 200
+    And match response.createdRecords == 1
+    And match response.updatedRecords == 0
+    And match response.failedRecords == 0
+    And match response.totalRecords == 1
 
+  Scenario: Verify user was imported successfully by getting the user and checking its properties
+    Given path 'users'
+    And param query = "barcode==" + barcode
```
julianladisch, line 62: The barcode string should be quoted.

```
    And param query = `barcode=="${barcode}"`
```

mod-user-import/src/main/resources/domain/mod-user-import/features/userImport.feature
```diff
@@ -19,36 +25,69 @@ Feature: User import
...
+    And assert response.users[0].barcode == barcode
+    And assert response.users[0].username == usersToImport.users[0].username
+    And assert response.totalRecords == 1
+    # Assign the user id for use in subsequent scenarios. The id is created when the user is inserted.
+    * def userId = response.users[0].id
+    * print userId
+
+  # Update the test users array with some new properties, including the id which will cause the update to happen.
+  Scenario: Import updated users with a changed JSON users array and check JSON response
+    * set usersToImport.users[0].id = userId;
```
julianladisch, line 75: Setting id should be removed. mod-user-import matches by externalSystemId.

mod-user-import/src/main/resources/domain/mod-user-import/features/userImport.feature
```diff
@@ -19,36 +25,69 @@ Feature: User import
...
+    When method POST
+    Then status 200
+    And match response.createdRecords == 0
+    And match response.updatedRecords == 1
+    And match response.failedRecords == 0
+    And match response.totalRecords == 1
+
+  Scenario: Verify JSON user update for a given user by checking the updated property
+    Given path 'users'
+    And param query = "barcode==" + barcode
```
julianladisch, line 88:
```suggestion
    And param query = `barcode=="${barcode}"`
```

mod-user-import/src/main/resources/domain/mod-user-import/features/userImport.feature
steveellis: Thanks Julian. Couldn't use the js template string but managed the same thing with concatenation.

mod-user-import/src/main/resources/domain/mod-user-import/features/userImport.feature
steveellis: This is no longer an issue since the update is happening using the externalSystemId.

hjiebsco (approved):

PR #206, opened 2021-06-30
FAT-169: Implement mod-login-saml API Karate tests based on test plans 

steveellis:
The ```orchestrate.feature``` is the top level of the tests. It creates an IdP with a POST binding and calls ```saml/login``` for it and does the same for a REDIRECT binding. In the spirit of integration testing, each configured IdP is an "actual test" IdP, taking suggestions for the two IdPs from the [guide](https://github.com/folio-org/mod-login-saml/blob/master/GUIDE.md).

I wasn't able to fully test the ```saml/callback``` route (what gets called by the IdP after logging in at the IdP), hopefully for obvious reasons, but there are some notes in the ```samlCallback.feature``` about that. Maybe there is a way to mock an "actual signature" from the IdP...? If so that requires more code fu than I have.

I think the rest assured tests cover more scenarios and edge cases but the coverage for the two binding types is new here (in addition to the tests running through okapi and a tenant).


steveellis:
Devoted a few brain cycles to this. Not loving that ```saml/callback``` isn't tested. I think it should be. There should be a way to simulate calling the endpoint from the IdP and get that 302 response.

The rest [assured test](https://github.com/folio-org/mod-login-saml/blob/master/src/test/java/org/folio/rest/impl/SamlAPITest.java#L386) just grabs the RelayState cookie from ```saml/login``` and posts it as a form param to ```saml/callback```. When I tried doing the same through karate I got a 500 error. That could be generated from [more than one place in the vertx route](https://github.com/folio-org/mod-login-saml/blob/f257300cae5207c5c8e46a01dbc16dfd25974b70/src/main/java/org/folio/rest/impl/SamlAPI.java#L158).

I thought it had something to do with the other form param ```SAMLResponse``` being signed by the IdP but looking more carefully at the [code](https://github.com/folio-org/mod-login-saml/blob/f257300cae5207c5c8e46a01dbc16dfd25974b70/src/main/java/org/folio/rest/impl/SamlAPI.java#L158) I'm not so sure about that. I don't see anywhere that the SAMLResponse param is being used at all, which would explain why the rest assured tests don't need it.

Part of the challenge here is that there are 5 or 6 places in the [method](https://github.com/folio-org/mod-login-saml/blob/f257300cae5207c5c8e46a01dbc16dfd25974b70/src/main/java/org/folio/rest/impl/SamlAPI.java#L158) that could be generating that 500 response, and none pass any details back in the body as far as I can see.

Does anybody have any ideas on this one?

What I would like to be able to do is deploy a build of mod-login-saml to the testing-backend that had more helpful logging on the monster method. Does anyone have a script for such things?

steveellis:
I may have been a little too optimistic about being able to test ```saml/callback```.  I spent more time looking into exactly what the IdP returns in the form of the SAMLResponse and don't see an easy way to create one that would pass validation. Here's what testsaml.id returns:

[SamlResponseExample.txt](https://github.com/folio-org/folio-integration-tests/files/6771308/SamlResponseExample.txt)

I could replace the tenant info with the test tenant, but it's unclear to me what would be considered valid values for the other items in the file.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/configureSaml.feature
```diff
@@ -0,0 +1,70 @@
...
+    # Here we need the idpUrl to actually be reachable and return XML.
+    And header Content-Type = "application/json"
+    And request
+    """
+    {
+      "idpUrl": "#(idpUrl)",
+      "samlBinding": "#(binding)",
+      "samlAttribute": "#(samlAttribute)",
+      "userProperty": "#(userProp)",
+      "okapiUrl": "http://localhost:9130"
```
julianladisch, line 37: Should use `config.baseUrl` configured in karate-config.js

mod-login-saml/src/main/resources/domain/mod-login-saml/features/configureSaml.feature
```diff
@@ -0,0 +1,70 @@
...
+      "X-Okapi-Token": "#(okapitoken)",
+      "Accept": "application/json, text/plain"
+    }
+    """
+  Scenario: Validate IdP url
+    Given path "saml/validate"
+    And param type = "idpurl"
+    And param value = idpUrl
+    When method GET
+    Then status 200
```
julianladisch, line 19: This scenario can be removed; is not needed for this kind of integration test. The GET saml/validate that checks the active flag in line 54 can be kept.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSamlRedirect.feature
```diff
@@ -0,0 +1,41 @@
...
+
+  Scenario: Do preflight request, POST to saml login endpoint, and receive correct response for REDIRECT binding
+    Given path "saml/login"
+    And header Origin = baseUrl
+    And header Access-Control-Request-Method = "POST"
+    When method OPTIONS
+    Then status 204
+    And match header access-control-allow-methods contains "POST"
+    And match header access-control-allow-origin == "*"
+    Given path "saml/login"
```
julianladisch, line 23: Can you add an empty line before `Given` for readability?

mod-login-saml/src/test/java/org/folio/ModLoginSamlTests.java
```diff
@@ -5,6 +5,7 @@
 import org.folio.test.services.TestIntegrationService;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
```
julianladisch, line 8:
```suggestion
```

mod-login-saml/src/main/resources/domain/mod-login-saml/features/samlFailures.feature
```diff
@@ -0,0 +1,46 @@
...
+  # I suspect this is because the karate tests are running through okapi. These responses seem correct though.
+  #
+
+  Scenario: Options request without origin fails with 404
+    Given path "saml/login"
+    And header Access-Control-Request-Method = "POST"
+    When method OPTIONS
+    Then status 404
+
+  Scenario: Options request with invalid fails with 403 bad request
```
julianladisch, line 40:
```suggestion
  Scenario: Options request with invalid Origin fails with 403 bad request
```

mod-login-saml/src/main/resources/domain/mod-login-saml/features/configureSaml.feature
steveellis: Agreed. Will remove.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/configureSaml.feature
steveellis: Agreed. Using the baseUrl would be better. There are a couple of other places where I use localhost for the okapi URL. I'll change those too.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSamlRedirect.feature
steveellis: Yes I can add that. Empty lines in scenarios don't appear to cause problems and it would make it easier to follow.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/samlFailures.feature
steveellis: Good catch. Thank you.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/samlCallback.feature
```diff
@@ -0,0 +1,26 @@
...
+
+  Scenario: Test CORS for the callback endpoint
+    Given path "saml/callback"
+    And header Origin = baseUrl
+    And header Access-Control-Request-Method = "POST"
+    When method OPTIONS
+    Then status 204
+    And match header access-control-allow-methods contains "POST"
+    And match header access-control-allow-origin == "*"
+    # The rest assured test checks here for Access-Control-Allow-Credentials = true. But this isn't returned
```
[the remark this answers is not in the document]
steveellis: I think it's registering it correctly. Let me know what you think. Here's the request that registers mod-login-saml with the test tenant that is generated on each test run:

```
10:25:08.509 request:
1 > POST https://folio-testing-okapi.dev.folio.org:443/_/proxy/tenants/testenant2379682412697761792/install?tenantParameters=loadSample%3Dfalse%2CloadReference%3Dfalse
1 > Accept: application/json
1 > Accept-Encoding: gzip,deflate
1 > Connection: Keep-Alive
1 > Content-Length: 61
1 > Content-Type: application/json; charset=UTF-8
1 > Host: folio-testing-okapi.dev.folio.org:443
1 > User-Agent: Apache-HttpClient/4.5.12 (Java/11.0.10)
1 > x-okapi-token: eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJ0ZXN0aW5nX2FkbWluIiwidXNlcl9pZCI6ImQ5ZDk1ODJlLTY2YWQtNWJkMC1iM2NiLTdkYjIwZTc1MzljYyIsImlhdCI6MTYyNTc1NDMwMywidGVuYW50Ijoic3VwZXJ0ZW5hbnQifQ.78HH8YffT_BkmORBfhwAJc3HOJ9d1uECShGS3-XRUXw
[
  {
    "id": "mod-login-saml-2.2.2-SNAPSHOT.73",
    "action": "enable"
  }
]


10:25:08.739 response time in milliseconds: 228.95
1 < 200
1 < Connection: keep-alive
1 < Content-Length: 74
1 < Content-Type: application/json
1 < Date: Thu, 08 Jul 2021 14:25:08 GMT
1 < X-Okapi-Trace: POST okapi-4.9.0-SNAPSHOT /_/proxy/tenants/testenant2379682412697761792/install.. : 200 98700us
1 < vary: origin
[
  {
    "id": "mod-login-saml-2.2.2-SNAPSHOT.73",
    "action": "enable"
  }
]
```
The id is obtained from this request:

```
10:25:08.369 request:
1 > GET https://folio-testing-okapi.dev.folio.org:443/_/proxy/modules?filter=mod-login-saml
1 > Accept: application/json
1 > Accept-Encoding: gzip,deflate
1 > Connection: Keep-Alive
1 > Content-Type: application/json
1 > Host: folio-testing-okapi.dev.folio.org:443
1 > User-Agent: Apache-HttpClient/4.5.12 (Java/11.0.10)
1 > x-okapi-token: eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJ0ZXN0aW5nX2FkbWluIiwidXNlcl9pZCI6ImQ5ZDk1ODJlLTY2YWQtNWJkMC1iM2NiLTdkYjIwZTc1MzljYyIsImlhdCI6MTYyNTc1NDMwMywidGVuYW50Ijoic3VwZXJ0ZW5hbnQifQ.78HH8YffT_BkmORBfhwAJc3HOJ9d1uECShGS3-XRUXw

10:25:08.504 response time in milliseconds: 133.55
1 < 200
1 < Connection: keep-alive
1 < Content-Length: 76
1 < Content-Type: application/json
1 < Date: Thu, 08 Jul 2021 14:25:08 GMT
1 < X-Okapi-Trace: GET okapi-4.9.0-SNAPSHOT /_/proxy/modules.. : 200 7131us
1 < vary: origin
[
  {
    "id": "mod-login-saml-2.2.2-SNAPSHOT.73",
    "name": "SAML login"
  }
]
```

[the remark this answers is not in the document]
hjiebsco: I know why. To trigger CORS delegate, the call to Okapi has to use `/_/invoke/tenant/{id}` format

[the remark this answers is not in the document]
julianladisch: @hjiebsco : We forgot to add OPTIONS to the module descriptor:
https://github.com/folio-org/mod-login-saml/blob/master/descriptors/ModuleDescriptor-template.json#L11-L22
@steveellis : Thank you for this good test, it caught this bug that we can only see when running an integration test that combines Okapi and the module.

julianladisch (approved):
Thank you, very good!!!

mod-login-saml/src/main/resources/domain/mod-login-saml/features/samlCallback.feature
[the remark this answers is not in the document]
hjiebsco: > @hjiebsco : We forgot to add OPTIONS to the module descriptor:
> https://github.com/folio-org/mod-login-saml/blob/master/descriptors/ModuleDescriptor-template.json#L11-L22
> @steveellis : Thank you for this good test, it caught this bug that we can only see when running an integration test that combines Okapi and the module.

I do not think that's the case. Okapi gets the initial option call and forward a call to the module. BTW this feature was tested in a deployed environment and it worked.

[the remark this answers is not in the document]
julianladisch: Regardless of the cause this needs to be fixed in mod-login-saml and requires a new mod-login-saml release.
Steve's pull request is ready to merge, we should not wait for the new mod-login-saml release.
If mod-login-saml has been fixed the integration test can be extended.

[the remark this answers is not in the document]
julianladisch: https://issues.folio.org/browse/MODLOGSAML-103 "CORS Access-Control-Allow-Credentials for /saml/login and /saml/callback"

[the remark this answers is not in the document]
hjiebsco: > Regardless of the cause this needs to be fixed in mod-login-saml and requires a new mod-login-saml release.
> Steve's pull request is ready to merge, we should not wait for the new mod-login-saml release.
> If mod-login-saml has been fixed the integration test can be extended.

I do not see the value of updating module descriptor with OPTIONS. But feel free to create a ticket and bring it to CP meeting. For this integration tests, I think we should test how FOLIO works. UI uses `invoke` to call Okapi and Okapi proxy requests to mod-login-saml.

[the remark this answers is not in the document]
hjiebsco: > https://issues.folio.org/browse/MODLOGSAML-103 "CORS Access-Control-Allow-Credentials for /saml/login and /saml/callback"

I added a comment to the ticket. Not sure if you remember but I believe the design was to delegate CORS only if called through invoke. Cc @adamdickmeiss and @craigmcnally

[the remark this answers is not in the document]
julianladisch: Yes, you are right and I am wrong.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/samlCallback.feature
```diff
@@ -0,0 +1,26 @@
...
+    * configure headers =
+    """
+    {
+      "X-Okapi-Tenant": "#(testTenant)",
+      "X-Okapi-Token": "#(okapitoken)"
+    }
+    """
+
+  Scenario: Test CORS for the callback endpoint
+    Given path "saml/callback"
```
hjiebsco, line 18:
```suggestion
    Given path "_/invoke/tenant/" + testTenant + "/saml/callback"
```

mod-login-saml/src/main/resources/domain/mod-login-saml/features/samlCallback.feature
```diff
@@ -0,0 +1,26 @@
...
+    And header Origin = baseUrl
+    And header Access-Control-Request-Method = "POST"
+    When method OPTIONS
+    Then status 204
+    And match header access-control-allow-methods contains "POST"
+    And match header access-control-allow-origin == "*"
```
hjiebsco, line 24:
```suggestion
    And match header access-control-allow-origin == baseUrl
    And match header access-control-allow-credentials == "true"
```

hjiebsco (commented):
Added in line suggestion for testing CORS handling. Other than that, LGTM.

mod-login-saml/src/main/resources/domain/mod-login-saml/features/samlCallback.feature
```diff
@@ -0,0 +1,25 @@
...
+
+  Scenario: Test CORS for the callback endpoint
+    Given path "_/invoke/tenant/" + testTenant + "/saml/callback"
+    And header Origin = baseUrl
+    And header Access-Control-Request-Method = "POST"
+    When method OPTIONS
+    Then status 204
+    And match header access-control-allow-methods contains "POST"
+    And match header access-control-allow-origin == baseUrl
+    And match header access-control-allow-credentials == "true"
```
steveellis, line 25: @hjiebsco and @julianladisch I think this should take care of it. To verify that things were working as expected based on Hongwei's comments, I added a SAML binding manually on the testing reference environment by going to Settings > Tenants, and indeed the route that is called by the client JS, after logging in at the IdP, is ```_invoke/tenant/<tenant id>/saml/callback```, and the the new revision of the test passes.

hjiebsco (approved):

mod-login-saml/src/main/resources/domain/mod-login-saml/features/loginSamlPost.feature
```diff
@@ -0,0 +1,48 @@
...
+    """
+    {
+      "X-Okapi-Tenant": "#(testTenant)",
+      "X-Okapi-Token": "#(okapitoken)",
+      "Accept": "application/json, text/plain"
+    }
+    """
+
+  Scenario: Do preflight request, POST to saml login endpoint, and receive correct response for POST binding
+    Given path "saml/login"
```
julianladisch, line 16: The actual OPTIONS request uses this uri: https://folio-snapshot-okapi.dev.folio.org/_/invoke/tenant/diku/saml/login

However, we should keep the test for the wrong path "saml/login" and check that the credentials header is NOT present:
```
  Background:
    * configure lowerCaseResponseHeaders = true
```

`And match responseHeaders contains { 'access-control-allow-credentials': '#notpresent' }`

And add a test for the correct path `"_/invoke/tenant/" + testTenant + "/saml/login"`

steveellis: Indeed you are right,  ```saml/login``` also goes through ```_/invoke/tenant/...```. I'm also noticing that the current test looks for the response header to be ```header access-control-allow-origin: "*"``` when I think it should be ```header access-control-allow-origin: <the origin url>```.

Noticing too that the ```saml/configure``` and perhaps other tests could have OPTIONS requests on them too. Will add those.

I also like the use of ```lowerCaseResponseHeaders``` and ```#notpresent``` on ```access-control-allow-credentials``` for a new test that checks the wrong path.

julianladisch (approved):
Well done

PR #207, opened 2021-07-01
FAT-8 - Karate tests for mod-audit

obozhko-folio:
[FAT-8](https://issues.folio.org/browse/FAT-8) - MODAUD: Implement API Karate tests

FAT-8-cucumber-report


mod-audit/src/main/resources/karate-config.js
```diff
@@ -67,10 +67,10 @@ function fn() {
     }
   } else if (env != null && env.match(/^ec2-\d+/)) {
     // Config for FOLIO CI "folio-integration" public ec2- dns name
-    config.baseUrl = 'http://' + env + ':9130';
+    config.baseUrl = 'https://folio-testing-okapi.dev.folio.org:443';
```
alekGbuz, line 70: For mod-audit testing previous value of baseUrl should work, why it was changed to testing endpoint?

obozhko-folio: It does not work for some reason:
karate_no_diku

mod-audit/src/main/resources/global/destroyTest.feature
```diff
@@ -41,7 +41,62 @@ Feature: remove all test data
...
 
+  @DeleteUserGroup
+  Scenario: Delete user group
+    Given path 'groups', userGroupId
+    When method DELETE
+
   @DeleteServicePoint
   Scenario: Delete user
     Given path 'service-points', servicePointId
+    When method DELETE
```
alekGbuz, line 52: There is execution of destroy-data.feature at tests. So it looks that it is not necessary to remove test data unless there is special case for such cleaning.

alekGbuz (approved):

mod-audit/src/main/resources/global/destroyTest.feature
obozhko-folio: Remove test data is required due to the invocation of initTest.feature before each scenario.

khandramai (approved):

siarhei-charniak (approved):

PR #214, opened 2021-07-09
[MODINVOICE-178] Check poNumbers updates when invoice lines are created and updated

damien-git:
## Purpose
[MODINVOICE-178](https://issues.folio.org/browse/MODINVOICE-178) - Update invoice.poNumbers field when invoiceLine is being linked to poLine


damien-git:
Screenshot_20210709_112143

PR #221, opened 2021-08-02
FAT-194 mod-password-validator: Created test plan

Shans-Kaluhin:
## Purpose
Adding undefined scenarios
## Learning
https://issues.folio.org/browse/FAT-194

### DIFF
--- a/mod-password-validator/src/test/resources/domain/passwordvalidator/features/validate.feature
+++ b/mod-password-validator/src/test/resources/domain/passwordvalidator/features/validate.feature
@@ -10,13 +10,41 @@ Feature: Test password validate
     * print 'undefined'
 
   @Undefined
-  Scenario: POST password should return 422 if validation error
+  Scenario: POST validate should return 422 if password contains consecutive whitespaces
+    * print 'undefined'
+
+  @Undefined
+  Scenario: POST validate should return 422 if password contains user name
+    * print 'undefined'
+
+  @Undefined
+  Scenario: POST validate should return 422 if password contains white space characters
+    * print 'undefined'
+
+  @Undefined
+  Scenario: POST validate should return 422 if password contains keyboard sequence
+    * print 'undefined'
+
+  @Undefined
+  Scenario: POST validate should return 422 if password contains repeating characters
+    * print 'undefined'
+
+  @Undefined
+  Scenario: POST validate should return 422 if password NOT contains special character
+    * print 'undefined'
+
+  @Undefined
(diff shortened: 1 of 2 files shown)


mod-password-validator/src/test/resources/domain/passwordvalidator/features/validate.feature
```diff
@@ -0,0 +1,22 @@
...
+  @Undefined
+  Scenario: POST password should return 422 if validation error
+    * print 'undefined'
+
+  @Undefined
+  Scenario: POST password should return 400 if bad request
+    * print 'undefined'
+
+  @Undefined
+  Scenario: POST password should return 500 if internal server error
```
psmagin, line 21: I am interested in how you will simulate 500 error

mod-password-validator/src/test/resources/domain/passwordvalidator/features/validate.feature
```diff
@@ -0,0 +1,22 @@
+Feature: Test password validate
+
+  Background:
+    * url baseUrl
+    * callonce login testUser
+    * configure headers = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapiToken)', 'Accept': 'application/json'  }
+
+  @Undefined
+  Scenario: POST validate should return 200 on success
```
psmagin, line 9: I think we should have several scenarios on 200:
- password is passing all enabled rules
- password is not passing a rule (for each default rule - separate scenario)

psmagin (changes requested):

mod-password-validator/src/test/resources/domain/passwordvalidator/features/validate.feature
Shans-Kaluhin: Didn't think about it, but I took this example from other modules

mod-password-validator/src/test/resources/domain/passwordvalidator/features/validate.feature
Shans-Kaluhin: Added new scenarios according to validation rules taken from our database

psmagin: Are you sure that if the password failed validation then we got 422 and not 200? As I see in API docs https://s3.amazonaws.com/foliodocs/api/mod-password-validator/s/password-validator.html it will be 200 for valid and not valid passwords.

psmagin (approved):

PR #237, opened 2021-08-10
FAT-934: Implement mod-permissions API Karate tests

steveellis:
This combines FAT-930, FAT-932 and FAT-934.

Tried to only put OPTIONs preflights on the main operations being tested, and not on requests that are there just verify that things went as planned.

This doesn't replicate all the tests from RestVerticleTest.java but instead tries to cover most of the routes used by the UI.


mod-permissions/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>mod-permissions</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <properties>
+        <cucumber.reporting.version>3.8.0</cucumber.reporting.version>
+        <testrail-api-java-client.version>2.0.1</testrail-api-java-client.version>
```
hjiebsco, line 19: Where/how are these properties used?

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
```diff
@@ -0,0 +1,398 @@
...
+    {
+      "Access-Control-Allow-Headers": "Content-Type, X-Okapi-Tenant, X-Okapi-Token, Authorization, X-Okapi-Request-Id, X-Okapi-Module-Id",
+      "Access-Control-Request-Method": "PUT, PATCH, DELETE, GET, POST",
+      "Access-Control-Allow-Origin": "*",
+      "Origin": "#(baseUrl)"
+    }
+    """
+    * configure lowerCaseResponseHeaders = true
+    * def testUserId = "00000000-1111-5555-9999-999999999998"
+    * def permissionsUserId = "00000000-1111-5555-9999-888888888885"
```
hjiebsco, line 26: Can those be generated or has to be static?

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
```diff
@@ -0,0 +1,398 @@
...
+
+  Scenario: Create permissions for a user and check that the permission has been granted
+    * def perms = ["okapi.all"]
+ 
+    # Do a preflight request to emulate the browser.
+    Given path 'perms/users'
+    And headers karate.merge(commonHeaders, optionsHeaders)
+    When method OPTIONS
+    Then status 204
+    And match header access-control-allow-methods contains "POST"
```
hjiebsco, line 58: In general, do we need test OPTIONS in specific module, or should it be covered in Okapi tests?

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
```diff
@@ -0,0 +1,398 @@
...
+  # so the existing test user (the one created in setup-users.feature) won't work because it already has
+  # permissions.
+  Scenario: Create a test user without any permissions yet
+    Given path 'users'
+    And headers commonHeaders
+    And request
+    """
+    {
+      "id": "#(testUserId)",
+      "username": "TestUser012",
```
hjiebsco, line 43: How to make sure `TestUser012` is not used by other tests?

hjiebsco (approved):
Curious, are `logback-test.xml` and `karate-config.js` same for all modules?

mod-permissions/pom.xml
steveellis: I think both of these lines can be safely removed. The cucumber reports still get generated by karate whether this reference is there. Only about half the projects include the testrail java client. May be needed depending on what you want to do, but after removing both, tests still run and reports still get generated.

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
steveellis: Agree this doesn't seem right. I got stuck because doing `*def id = uuid()` was generating a new uuid everytime the variable was used (it was a function reference in other words rather than the value). But if you do `* def  id = callonce uuid` things work as expected. I'll make the change.

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
steveellis: I think the OPTIONS requests are useful here. Including them uncovered interesting issues in our mod-login-saml tests for example. I think it's good to try to emulate what browser does as much as we can.

steveellis:
@hjiebsco I think `logback-test.xml` and `karate-config.js` are mostly the same between projects. I think either one could be customized when needed depending on the project. For example, on the test for mod-user-import it was convenient to include the test user data to import in the karate-config.js file since other test specific data is stored there. Other tests could add their own custom functions there depending on their needs.

hjiebsco (approved):
Looks good to me. Thanks.

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
julianladisch: Having OPTIONS integration tests for one module is fine, and for APIs where CORS is delegated from Okapi to the module, but not for all modules.

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
```diff
@@ -0,0 +1,397 @@
...
+  When method OPTIONS
+  Then status 204
+  And match header access-control-allow-methods contains "POST"
+
+  Given path 'perms/users/', permissionsUserId, 'permissions'
+  And headers commonHeaders
+  And request { permissionName: #(permToAdd) }
+  When method POST
+  Then status 200
+  And match response == { permissionName: #string }
```
julianladisch, line 175: Isn't this string check already covered by the following line?

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
```diff
@@ -0,0 +1,397 @@
...
+  Then status 200
+  And match response.permissions[0].grantedTo !contains permissionsUserId
+
+  #
+  # Test operations on permissions themselves. This is something that happens in the
+  # browser under Settings > Users > Permission sets.
+  #
+
+  Scenario: Get a certain number of permissions for the tenant and validate the response
+    * def numberToGet = 10
```
julianladisch, line 225: This is the default. Better use a non-default number, for example 12.

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
```diff
@@ -0,0 +1,397 @@
...
+      "childOf": "#array",
+      "grantedTo": "#array",
+      "subPermissions": "#array",
+      "dummy": "#boolean",
+      "mutable": "#boolean",
+      "visible": "#boolean",
+      "deprecated": "#boolean",
+      "metadata": "##object"
+    }
+    """
```
julianladisch, line 260: We don't want to update the integration test when we add new properties to the permission record. As long as other modules can successfully handle the new property that integration test should succeed.

steveellis: I could be persuaded either way. If we do no schema validation then we won't catch (however unlikely) situations where an important property stops getting included. How about if we define a few properties as critical and just check for those? Good candidates might be `id`, `childOf`, `grantedTo`, `subPermissions`. 

I think the karate way of doing that would be a little more verbose but still useful:

```
And match response[0].childOf == #array
```

There probably would be a way to do it in the context of the `each` which seems like a useful construct.

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
steveellis: Yes the type check is a bit redundant.

mod-permissions/src/main/resources/domain/mod-permissions/features/permissions.feature
steveellis: ` And match each response contains { "property: "#string" }` seems like the way to validate parts of a schema using `each`.

julianladisch: To catch that module A stops including an important property we write an integration test where module B fails when that property is missing. Module B can be a back-end or front-end module.

The test whether module A returns property X is a unit test, not an integration test.

steveellis: Thanks for clarifying. I like the approach. Made the changes you recommended.

julianladisch (approved):
Looks good, thank you!

PR #244, opened 2021-08-13
FAT-933: Implement karate tests for mod-users-bl and mod-login

steveellis:
This should resolve FAT-933, FAT-929, and FAT-931. It may also contribute to FAT-161, FAT-165 and FAT-175 - the last three being tickets for mod-users-bl. The first three being tickets for mod-login.

**Background**
As I began to work on mod-login, I noticed that all UI requests were first flowing through mod-users-bl, and mod-login was only being called by mod-users-bl. After discussing with the CP team we agreed that doing karate tests for mod-login wouldn't be useful to us, but that I should try to contribute some tests to mod-users-bl. That's what this PR is about. Once this PR is merged, the team with responsibility for mod-users-bl will have a good start on their karate stuff. 

With all this in mind I'm inviting members of the CP team and contributors to mod-users-bl to review this PR.

For those unfamiliar with our Karate setup see [this](https://wiki.folio.org/pages/viewpage.action?spaceKey=FOLIJET&title=Karate+API+Integration+Testing). To run the tests for this module, grab this branch and run
`./runtests.sh mod-users-bl testing` in your terminal. This will create a tenant, register the relevant modules, create some users, run the tests against the testing reference env, and then clean up after itself.

One way to view the test results is to see the surefire-reports directory in the sub project's (in this case mod-users-bl) target.


julianladisch (approved):
The invalid password and wrong username scenarios are not needed because they test the same integration that is already covered by the scenario with successful login.

hjiebsco (approved):

steveellis:
@julianladisch Thanks for trying to sharpen my understanding on this. I do appreciate it! Indeed the two additional tests seem to be covered in the unit tests in places like [this](https://github.com/folio-org/mod-users-bl/blob/c9cd3b8701f6744f2390cdf4354d41c0f12da0cd/src/test/java/org/folio/rest/GeneratePasswordRestLinkTest.java#L325).

mod-users-bl/src/test/java/org/folio/ModUsersBLTests.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @BeforeAll
+  public void setup() {
+    runFeature("classpath:domain/mod-users-bl/users-bl-junit.feature");
+  }
+
+  @AfterAll
+  public void tearDown() {
+    runFeature("classpath:common/destroy-data.feature");
+  }
```
marcjohnson-kint, line 25: Does this mean that after all of the tests in this test are run, all of the data set up by the common part of the Karate tests are deleted?

marcjohnson-kint: There doesn't seem to be a corresponding common function for creating data. What triggers the creation of the tenant etc?

mod-users-bl/src/main/resources/domain/mod-users-bl/features/users-bl.feature
```diff
@@ -0,0 +1,79 @@
...
+    }
+    """
+
+    # Login the test user. This user was created in common/setup-users.feature.
+    Given path 'bl-users/login'
+    And header Origin = baseUrl
+    And header Access-Control-Request-Method = "POST"
+    When method OPTIONS
+    Then status 204
+    And match header access-control-allow-methods contains "POST"
```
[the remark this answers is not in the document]
steveellis: I think it's debatable whether we should do this. My argument has been that a goal should be to emulate the browser as much as we can and make sure that preflights return the right responses.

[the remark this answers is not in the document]
marcjohnson-kint: > My argument has been that a goal should be to emulate the browser as much as we can and make sure that preflights return the right responses.

From that perspective, making an OPTIONS request is a reasonable thing to do.

My understanding of the intent of these tests is not to try to capture every detail of how the browser running FOLIO's UI makes requests and instead to demonstrate that the important parts of it are working as expected.

I think there is a trade off between having the most thorough and complete set of tests at this level and how expensive this style of test is to write, run and maintain.

mod-users-bl/src/test/java/org/folio/ModUsersBLTests.java
steveellis: I think the call chain is:

1. JUnit invokes the setup method in ModUsersBLTests.java which calls users-bl-junit.feature
2. users-bl-junit.feature calls common/setup-users.feature
3. setup-users.feature calls common/tenant.feature@create etc

One nice thing is that the each module runs in its own space, creating its own tenant, registering the modules it needs, and then cleans up after itself. So there's no danger of collisions between tests.

steveellis: When I say "each module" I mean each sub-project in the main repo. The modules that get registered for the tenant that gets created are defined in users-bl-junit.feature.

marcjohnson-kint: > When I say "each module" I mean each sub-project in the main repo.

I understand that. This confusion of terminology is caused by the decision to organise these tests by module rather than capability.

marcjohnson-kint: > 1. JUnit invokes the setup method in ModUsersBLTests.java which calls users-bl-junit.feature
> 2. users-bl-junit.feature calls common/setup-users.feature
> 3. setup-users.feature calls common/tenant.feature@create etc
 
Thank you for this explanation.

What I don't think I understand is the asymmetry. The setup code is defined in the feature, yet the tear down code is defined in the JUnit test.

> So there's no danger of collisions between tests.

I think you mean between tests in other sub-projects. Tests in the same sub-project could presumably clash with each other because their share state.

steveellis: Theres a setup() method in `ModPermsisionsTests.java` that controls the setup too. So it's all flowing out of the java file, setup and teardown. So I guess it's symmetrical in that sense. The organization of the features is another matter though. It's really up to us to determine that for our sub-project, but otherwise I'm taking what has been given (the organization of stuff in `common`).

steveellis:
@marcjohnson-kint I can't seem to reply to your message about about where the data gets created. The `common/setup-users.feature` is where the users are created and it calls a sub feature that creates the tenants. `common/setup-users.feature` is called by our `...-junit.feature` in the setup() function in the `ModPermissionsTest.java` java class.

mod-users-bl/src/test/java/org/folio/ModUsersBLTests.java
marcjohnson-kint: > Theres a setup() method in `ModPermsisionsTests.java` that controls the setup too. So it's all flowing out of the java file, setup and teardown. 

Did you mean `ModUsersBLTests.java`?


> So I guess it's symmetrical in that sense. The organization of the features is another matter though. It's really up to us to determine that for our sub-project, but otherwise I'm taking what has been given (the organization of stuff in `common`).

I understand taking an approach that fits with the information you've been given. I'm playing catch up and trying to understand it. That may well be an unfair burden to put on you for offering to help. 

(Within the FOLIO community, I'm also a bit of a radical when it comes to my beliefs about software development)

marcjohnson-kint:
@steveellis Your team has given you approval to merge this work. 

I don't have any feedback specific to this work that should stop that (beyond my feedback about the options request). 

Are you waiting for my approval to merge this?

steveellis:
No problem Marc. I was hoping to get a green check mark from you if you don't mind.

marcjohnson-kint (approved):

PR #252, opened 2021-08-24
[MODORDERS-539] Check encumbrances after a fund distribution is removed

damien-git:
## Purpose
[MODORDERS-539](https://issues.folio.org/browse/MODORDERS-539) - Encumbrance not deleted when order Fund distribution is removed


damien-git:
Screenshot_20210824_110733

AndreiMPV (approved):

PR #289, opened 2021-09-28
[MODORDERS-568] Piece operations

damien-git:
## Purpose
[MODORDERS-568](https://issues.folio.org/browse/MODORDERS-568) - Create/update/merge API tests for covering functionality of the storing holding reference in the POL or piece

## Approach
The new test checks many operations and cases related to orders and pieces.

Some tests do not pass yet, as issues have to be resolved. In particular:
- Location quantities have to be updated when a piece is created
- Holdings should be preserved during unopen
- After an open order line is deleted, items that have the "on order" status need to be removed

## TODO
**Need to revise cases again, because almost of them was covered in the another features.
Also need better to split feature between package and non-package**
Story created https://issues.folio.org/browse/FAT-1106


AndreiMPV (approved):

AndreiMPV:
image

PR #302, opened 2021-10-11
FAT-143: Setup karate tests for mod-inventory

SinghAdes:


mod-inventory/src/main/resources/domain/mod-inventory/features/inventoryFeatureTest.feature
```diff
@@ -0,0 +1,12 @@
+Feature: inventory sample
+
+  Background:
+    * url baseUrl
+    * callonce login testUser
+    * configure headers = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': 'application/json, text/plain' }
+    
+
+  Scenario: print
```
steveellis, line 9: This should have an `@Undefined` annotation as should any further "test stubs" (specified but unimplemented tests). I believe adding this annotation to scenarios will allow the scenario to appear in the Cucumber reports as undefined (i.e. work which needs to be done) which is useful to the team overseeing the karate tests.

steveellis: Here's a link in the docs about the use of this `@Undefined` annotation: https://wiki.folio.org/display/FOLIJET/Karate+test+plan+for+cucumber+report

steveellis (approved):
Looks good, thanks. Going forward, may not want to put `_` in method names. I believe that's unconventional for this project (at least I can't recall seeing it being done).

PR #312, opened 2021-10-18
Modinv 509 create karate test for mod inventory

SinghAdes:


mod-inventory/src/main/resources/domain/mod-inventory/inventory-junit.feature
```diff
@@ -4,16 +4,33 @@ Feature: mod-inventory integration tests
...
+      |'inventory-storage.instances.item.get'                            |
+      |'inventory-storage.preceding-succeeding-titles.collection.get'    |
+      |'inventory-storage.preceding-succeeding-titles.item.post'         |
+      |'inventory-storage.preceding-succeeding-titles.item.put'          |
+      |'inventory-storage.preceding-succeeding-titles.item.delete'       |
+      |'inventory-storage.instance-relationships.collection.get'         |
+      |'inventory-storage.instance-relationships.item.post'              |
+      |'inventory-storage.instance-relationships.item.put'               |
+      |'inventory-storage.instance-relationships.item.delete'            |
+      |'inventory.instances.item.post'                                   |
```
marcjohnson-kint, line 33: Why does this test need this many permissions when it is created an instance via the API, which should only require  `inventory.instances.item.post`?

PR #321, opened 2021-10-21
* MODINV-512 to MODINV-518 -create Karate tests for MOD-INVENTORY.

SinghAdes:
@steveellis :
Task:-Create Karate test for Inventory: When duplicating an existing holdings, then a new Holdings is created with a unique UUID and Holdings HRID (inv-5).

Here i don't think generating id by calling UUID function is good. kindly let me know if any suggestion.

### DIFF
--- a/mod-inventory/src/main/resources/domain/mod-inventory/features/inventoryFeatureTest.feature
+++ b/mod-inventory/src/main/resources/domain/mod-inventory/features/inventoryFeatureTest.feature
@@ -50,3 +50,180 @@ Feature: inventory sample
       """
       When method POST
       Then status 201
+
+    Scenario: duplicate Instance,Holdings,Item creation
+#     First Instance
+      Given path 'inventory/instances'
+      And request
+      """
+      {
+        "source":"FOLIO",
+        "title":"TestInstance",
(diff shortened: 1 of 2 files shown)


mod-inventory/src/main/resources/domain/mod-inventory/features/inventoryFeatureTest.feature
```diff
@@ -50,3 +50,180 @@ Feature: inventory sample
...
+        "title":"TestInstance",
+        "instanceTypeId":"6312d172-f0cf-40f6-b27d-9fa8feaf332f"
+      }
+      """
+      When method POST
+      Then status 201
+      * def dupInstancelocation = responseHeaders['Location'][0]
+      * def dupInstanceId = dupInstancelocation.substring(dupInstancelocation.lastIndexOf('/') + 1)
+#     Both instances ids should not match
+      * match dupInstanceId != instanceId
```
marcjohnson-kint, line 85: Why is the instance being duplicated when the description specifically refers to duplicating only a holdings?

SinghAdes: @marcjohnson-kint :
We have Tasks for Instance & item also. 
i put that just to refer that if we use UUID function to generate ID , that wont justify the TASK .

PR #342, opened 2021-11-08
UXPROD-3138-create karate tests for mod-users.

SinghAdes:


mod-users/src/test/java/org/folio/ModUsersTests.java
```diff
@@ -0,0 +1,32 @@
...
+import org.folio.test.TestBase;
+import org.folio.test.config.TestModuleConfiguration;
+import org.folio.test.services.TestIntegrationService;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class ModUsersTests extends TestBase {
+
+  private static final String TEST_BASE_PATH = "classpath:domain/mod-users/features/";
```
marcjohnson-kint, line 12: This is only used in one place, why is it a class level constant?

mod-users/src/main/resources/logback-test.xml
```diff
@@ -0,0 +1,24 @@
...
+    </appender>
+
+    <logger name="com.intuit.karate" level="DEBUG"/>
+
+    <root level="info">
+        <appender-ref ref="STDOUT"/>
+        <appender-ref ref="FILE"/>
+    </root>
+
+</configuration>
```
marcjohnson-kint, line 24: Is this needed for every sub-project?

mod-users/src/test/java/org/folio/ModUsersTests.java
```diff
@@ -0,0 +1,32 @@
...
+    runFeature("classpath:domain/mod-users/users-junit.feature");
+  }
+
+  @AfterAll
+  public void tearDown() {
+    runFeature("classpath:common/destroy-data.feature");
+  }
+
+  @Test
+  void loanTest() {
```
marcjohnson-kint, line 29: Why is this called `loanTest` when it is a set of tests about users?

mod-users/src/main/resources/domain/mod-users/features/util/initData.feature
```diff
@@ -0,0 +1,30 @@
...
+    * url baseUrl
+    * callonce login testUser
+    * configure headers = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': 'application/json, text/plain' }
+    * def patronId = call uuid1
+    * def patronName = call random_string
+    * def barCode = call random_numbers
+
+  @PostPatronGroupAndUser
+  Scenario: create PatronGroup & User
+    * def createPatronGroupRequest = read('samples/PatronGroup/create-patronGroup-request.json')
```
marcjohnson-kint, line 13: I thought @steveellis suggested in a previous pull request that these kinds of uses are better inlined in the code.

@steveellis Am I mis-recollecting feedback you gave previously?

mod-users/src/main/resources/domain/mod-users/users-junit.feature
```diff
@@ -0,0 +1,28 @@
+Feature: mod-users integration tests
+
+  Background:
+    * url baseUrl
+    * table modules
+      | name                      |
+      | 'mod-login'               |
+      | 'mod-permissions'         |
+      | 'mod-users'               |
+      | 'mod-circulation-storage' |
```
marcjohnson-kint, line 10: Why is mod-circulation-storage needed for a users test?

mod-users/src/main/resources/karate-config.js
```diff
@@ -0,0 +1,87 @@
+function fn() {
```
marcjohnson-kint, line 1: Is this a copy and paste of the config from another test project?

mod-users/src/main/resources/domain/mod-users/features/util/initData.feature
SinghAdes: we need an **USER** for almost every test case, so writing multiple lines of same code for each test case is i don't think good way .
we have annotation facility also in karate to reduce the lines of code.
 Let me know if any suggestion on this.

mod-users/src/main/resources/domain/mod-users/users-junit.feature
SinghAdes: i will correct it . thankyou

mod-users/src/main/resources/karate-config.js
SinghAdes: Yes config file is same for all projects except 1 method which i have added for generating unique barcode.

**random_numbers: function() {
      var text = Math.floor(Math.random() * 10000);
      return text;
    }**

mod-users/src/test/java/org/folio/ModUsersTests.java
SinghAdes: Yes i miss this , i will correct it .

mod-users/src/main/resources/domain/mod-users/features/util/initData.feature
marcjohnson-kint: > we need an **USER** for almost every test case

I haven't questioned the need at all.

> so writing multiple lines of same code for each test case is i don't think good way 

Isn't this file already a definition of shared functions that are invoked from other places or have I misinterpreted what this is?

> we have annotation facility also in karate to reduce the lines of code. Let me know if any suggestion on this.

What annotation facility are you referring to?

mod-users/src/main/resources/logback-test.xml
SinghAdes: Yes i think . 
this file is needed for generating **KARATE.log** file.

mod-users/src/main/resources/domain/mod-users/features/util/initData.feature
marcjohnson-kint: I'm closing this following a conversation we had privately. 

I don't think we are going to reach an agreement on how to define the JSON in these tests.

mod-users/src/test/java/org/folio/ModUsersTests.java
SinghAdes: i am not exactly sure about it but i guess because of TestBase class constructor, since in all other projects it is same so i have kept as it is.

marcjohnson-kint: I don't understand. 

How could a (non-overriding) variable defined in the sub-class have any impact on what is being done in the super-class?

mod-users/src/test/java/org/folio/ModUsersTests.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  @AfterAll
+  public void tearDown() {
+    runFeature("classpath:common/destroy-data.feature");
+  }
+
+  @Test
+  void usersTest() {runFeatureTest("users");
+  }
```
marcjohnson-kint, line 30:
```suggestion
  @Test
  void usersTest() {
    runFeatureTest("users");
  }
```

mod-users/src/main/resources/domain/mod-users/features/users.feature
```diff
@@ -0,0 +1,111 @@
...
+
+
+
+
+
+
+
+
+
+
```
marcjohnson-kint, line 111: Why is there a bunch of white space at the bottom of this file?

mod-users/src/main/resources/domain/mod-users/features/users.feature
```diff
@@ -0,0 +1,111 @@
...
+    Given path 'users?query=(username=test-user)'
+    When method GET
+    Then status 200
+
+  Scenario: Filter inactive patron.
+    * call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser') { status: true }
+    * call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser') { status: true }
+    * call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser') { status: false }
+    * call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser') { status: false }
```
marcjohnson-kint, line 95: This filter scenario defines both matching and non-matching data to demonstrate the criteria is applied as expected.

The search scenarios don't seem to do this. They only define the matching data.

Please can you help me understand why this is done for this scenario and not the others?

mod-users/src/main/resources/domain/mod-users/features/util/initData.feature
```diff
@@ -0,0 +1,25 @@
...
+    Given path 'groups'
+    And request createPatronGroupRequest
+    When method POST
+    Then status 201
+    * def patronGroupId = response.id
+
+    * def createUserRequest = read('samples/User/create-user-request.json')
+    * createUserRequest.active = karate.get('status')
```
marcjohnson-kint, line 21: What does `karate.get` do? Is this how parameters are expressed in these executable chunks of code?

mod-users/src/main/resources/domain/mod-users/features/samples/PatronGroup/create-patronGroup-request.json
```diff
@@ -0,0 +1,6 @@
+{
+"group":"#(patronName)",
+"desc":"For Testing",
+"expirationOffsetInDays":"60",
+"id": "#(patronId)"
+}
```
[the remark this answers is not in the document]
SinghAdes: sure, i will make the changes.

mod-users/src/main/resources/domain/mod-users/features/users.feature
```diff
@@ -0,0 +1,110 @@
...
+    * def createProxyUserResponse = call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser')
+    * def proxyUserId = createProxyUserResponse.response.id
+    Given path 'proxiesfor'
+    And request {"accrueTo":"Sponsor","notificationsTo":"Sponsor","requestForSponsor":"Yes","status":"Active","proxyUserId":"#(proxyUserId)","userId":"#(userId)"}
+    When method POST
+    Then status 201
+    And match proxyUserId == response.proxyUserId
+
+  Scenario: Search user by firstname.
+    * def expectedFirstName = 'first'
```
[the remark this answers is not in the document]
SinghAdes: Yes , that is my mistake , i will correct it .thankyou.

mod-users/src/main/resources/domain/mod-users/features/users.feature
SinghAdes: i will remove them , those are automatically saved .

mod-users/src/main/resources/domain/mod-users/features/users.feature
SinghAdes: Yes i agree, 
i think in search scenario both matching and non-matching criteria should be there. i will correct it.  thankyou

mod-users/src/main/resources/domain/mod-users/features/util/initData.feature
SinghAdes: Karate.get() is used to get  value safe mode, but please ignore that i added for something to test and forgot to remove , i will make correction in next commit. thankyou.

mod-users/src/test/java/org/folio/ModUsersTests.java
```diff
@@ -0,0 +1,33 @@
...
+  public void tearDown() {
+
+    runFeature("classpath:common/destroy-data.feature");
+  }
+
+  @Test
+  void usersTest() {
+
+  runFeatureTest("users");
+  }
```
marcjohnson-kint, line 32:
```suggestion
  @Test
  void usersTest() {
    runFeatureTest("users");
  }
```

mod-users/src/test/java/org/folio/ModUsersTests.java
```diff
@@ -0,0 +1,33 @@
...
+  public void setup() {
+
+    runFeature("classpath:domain/mod-users/users-junit.feature");
+  }
+
+  @AfterAll
+  public void tearDown() {
+
+    runFeature("classpath:common/destroy-data.feature");
+  }
```
marcjohnson-kint, line 26: Personally, I don't think the extra blank lines improve the readability

```suggestion
  @BeforeAll
  public void setup() {
    runFeature("classpath:domain/mod-users/users-junit.feature");
  }

  @AfterAll
  public void tearDown() {
    runFeature("classpath:common/destroy-data.feature");
  }
```

mod-users/src/main/resources/domain/mod-users/features/users.feature
```diff
@@ -0,0 +1,151 @@
...
+    * def username = call random_string
+    * def createUserResponse = call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser') { barcode: 2222}
+    * def uuid = call uuid1
+    * def username = call random_string
+    * def createUserResponse = call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser') { barcode: 3333}
+
+    Given path 'users?query=(barcode=2222)'
+    When method GET
+    Then status 200
+    And match response.users[0].barcode == '2222'
```
marcjohnson-kint, line 30: This only checks that the first user found has the expected barcode. It doesn't seem to check that no other users are found, meaning this would pass if the second user was also found.

Maybe a check on the count of records found could avoid this.

mod-users/src/main/resources/domain/mod-users/features/users.feature
```diff
@@ -0,0 +1,151 @@
...
+    * def uuid = call uuid1
+    * def username = call random_string
+    * def barcode = call random_numbers
+    * def createUserResponse = call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser') { firstName: xyz }
+    * def responseBarcode = createUserResponse.response.firstName
+
+    Given path 'users?query=(personal.firstName=abc)'
+    When method GET
+    Then status 200
+    And match response.users[0].personal.firstName == 'abc'
```
marcjohnson-kint, line 64: Similar the above, this only checks that the first user found has the expected barcode. It doesn't seem to check that no other users are found, meaning this would pass if the second user was also found. 

Maybe a check on the count of records found could avoid this.

SinghAdes: @marcjohnson-kint 
Yes i agree, thankyou.

mod-users/src/main/resources/domain/mod-users/features/users.feature
```diff
@@ -0,0 +1,158 @@
...
+    * def lastName = call random_string
+    * def firstName = call random_string
+    * def username = call random_string
+    * def email = 'abc@pqr.com'
+
+  Scenario: Create a new User with PatronGroup.
+    * def username = call random_string
+    * def barcode = call random_numbers
+    * def uuid = call uuid1
+    * def createUserResponse = call read('classpath:domain/mod-users/features/util/initData.feature@PostPatronGroupAndUser')
```
marcjohnson-kint, line 17: Why define a variable to capture the response when the test does not inspect it in any way?

(similarly for the similar examples below)

marcjohnson-kint (approved):

PR #345, opened 2021-11-11
FAT-1003 - Created Karate test for Loans

sherzodn:
## Purpose
[FAT-1003](https://issues.folio.org/browse/FAT-1003)
_Create Karate test for Loans: When an loaned item is checked in at a service point that serves its location and no request exists, change the item status to Available (circ-13)_

## Approach

## Learning
- learn karate framework from original doc file on the github
- looked other scenarios how they are described.

### DIFF
--- a/mod-circulation/src/main/resources/domain/mod-circulation/features/loans.feature
+++ b/mod-circulation/src/main/resources/domain/mod-circulation/features/loans.feature
@@ -219,13 +219,13 @@ Feature: Loans tests
 
   Scenario: When an loaned item is checked in at a service point that serves its location and no request exists, change the item status to Available
 
-    * def extItemBarcode = '' + random(100000)
+    * def extItemBarcode = 'fat1003-ibc'
     * def extUserId = call uuid1
-    * def extUserBarcode = random(1000000)
+    * def extUserBarcode = 'fat1003-ubc'
 
     # location and service point setup
     * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostLocation')
-    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostServicePoint') { id: #(servicePointId) }
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostServicePoint')
 
     # post an item
     * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostInstance')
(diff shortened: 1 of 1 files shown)


mod-circulation/src/main/resources/domain/mod-circulation/features/loans.feature
```diff
@@ -216,3 +216,38 @@ Feature: Loans tests
     Then status 200
     Then match $.loans[0].dueDateChangedByRecall == true
     And match $.loans[0].dueDate !contains expectedDueDateBeforeRequest
+
+  Scenario: When an loaned item is checked in at a service point that serves its location and no request exists, change the item status to Available
+
+    * def extItemBarcode = '' + random(100000)
```
alexanderkurash, line 222: Is it necessary to use random values instead of constants?

mod-circulation/src/main/resources/domain/mod-circulation/features/loans.feature
```diff
@@ -216,3 +216,38 @@ Feature: Loans tests
...
+    * def extUserId = call uuid1
+    * def extUserBarcode = random(1000000)
+
+    # post an item
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostInstance')
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostServicePoint') { id: #(servicePointId) }
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostLocation')
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostHoldings')
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostItem') { extItemId: #(itemId), extItemBarcode: #(extItemBarcode)}
```
alexanderkurash, line 231: Can we reorder these or group them differently? Locations and service points are related to the institution structure but instances, holdings and items are more about inventory.

mod-circulation/src/main/resources/domain/mod-circulation/features/loans.feature
```diff
@@ -216,3 +216,38 @@ Feature: Loans tests
...
+
+    # post an user
```
alexanderkurash, line 233: "a user"?

mod-circulation/src/main/resources/domain/mod-circulation/features/loans.feature
```diff
@@ -216,3 +216,38 @@ Feature: Loans tests
...
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostPolicies')
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostGroup')
+    * call read('classpath:domain/mod-circulation/features/util/initData.feature@PostUser') { extUserBarcode: #(extUserBarcode), extUserId: #(extUserId) }
+
+    # checkOut the created item
```
alexanderkurash, line 238: Here and also on lines 241, 249 - let's keep the naming consistent, either call it a "check out" and "check in" or "check-out" and "check-in"

mod-circulation/src/main/resources/domain/mod-circulation/features/loans.feature
sherzodn: not necessary of course, but as far as I know, Viktor suggested using random values in previous reviews. So, what should I do?

sherzodn: @alexanderkurash done

PR #351, opened 2021-11-16
FAT-1011: When an item that had the status of in process that was checked out is checked in, set the item status to Available

yusuf-murodov-epam:
## Purpose
[FAT-1011](https://issues.folio.org/browse/FAT-1011)
_Create Karate test for Loans: When an item that had the status of in process that was checked out is checked in, set the item status to Available_

## Approach

## Learning
- learn karate framework from original doc file on the github
- looked other scenarios how they are described.

### DIFF
--- a/acquisitions/src/main/resources/domain/mod-orders/features/should-decrease-quantity-when-delete-piece-with-no-location.feature
+++ b/acquisitions/src/main/resources/domain/mod-orders/features/should-decrease-quantity-when-delete-piece-with-no-location.feature
@@ -1,0 +1,211 @@
+@parallel=false
+Feature: Should decrease quantity when delete piece with no location
+
+  Background:
+    * url baseUrl
+    # uncomment below line for development
+    #* callonce dev {tenant: 'test_orders1'}
+    * callonce login testAdmin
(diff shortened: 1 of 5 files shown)


viktor-draban (approved):

acquisitions/src/main/resources/domain/mod-orders/features/should-decrease-quantity-when-delete-piece-with-no-location.feature
```diff
@@ -0,0 +1,211 @@
+@parallel=false
+Feature: Should decrease quantity when delete piece with no location
```
alexanderkurash, line 2: Why are there changes in mod-orders tests?

alexanderkurash (changes requested):

acquisitions/src/main/resources/domain/mod-orders/features/should-decrease-quantity-when-delete-piece-with-no-location.feature
yusuf-murodov-epam: This occurs when I rebase with masters

PR #366, opened 2021-11-22
[MODORDERS-616] Test changing location when receiving a piece

damien-git:
## Purpose
[MODORDERS-616](https://issues.folio.org/browse/MODORDERS-616) - New holdingId is not saved into piece when location is changed during receiving

## Approach
Added a new test for this case, checking the piece's holding id is the one of the holding with the new location.


grabsefx:
test succesfull on folio testing
image

PR #367, opened 2021-11-24
UXPROD-3394-Karate-Test-for-EDGE-RTAC

SinghAdes:


edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
```diff
@@ -0,0 +1,148 @@
...
+    * def instanceId = call random_uuid
+    * def servicePointId = call random_uuid
+    * def locationId = call random_uuid
+    * def holdingId = call random_uuid
+    * def itemStatusName = 'Available'
+    * def periodicalStatus = true
+    * def materialTypeId = call random_uuid
+    * callonce read('classpath:domain/edge-rtac/features/util/initData.feature@PostMaterialType')
+
+  Scenario: For a non-periodical/non-serial, return holdings and item information including availability for an instance UUID
```
marcjohnson-kint, line 16: What about this scenario is specific to a non-periodical/non-serial?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
```diff
@@ -0,0 +1,148 @@
...
+    * def createSecondItemResponse = call read('classpath:domain/edge-rtac/features/util/initData.feature@PostItem') { itemId: 68640328-788e-53fc-9c3c-af39e243f3b7 }
+    * def periodicalStatus = true
+    * def rtacBatchrequest = read('samples/rtac-batch-request-entity.json')
+
+    Given path 'rtac-batch'
+    And request rtacBatchrequest
+    When method POST
+    Then status 200
+    And match response.holdings..id contains [ 68640328-799e-43fc-9c3c-af39e243f3b7 , 68640328-788e-53fc-9c3c-af39e243f3b7 ]
+    And match response.holdings..status contains [ Available , Checked out ]
```
marcjohnson-kint, line 64: Does that match applying ordering?

If it doesn't, is it important that the status is for the right holdings?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
```diff
@@ -0,0 +1,148 @@
...
+
+  Scenario:For periodical/serial, return holdings and item information including availability for each instance UUID included in request WHEN &fullPeriodicals=true
```
marcjohnson-kint, line 66: What about this scenario is specific to periodicals?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
```diff
@@ -0,0 +1,148 @@
...
+    * def secondInstanceId = createInstanceResponse.instanceEntityRequest.id
+    * call read('classpath:domain/edge-rtac/features/util/initData.feature@PostServicePoint')
+    * call read('classpath:domain/edge-rtac/features/util/initData.feature@PostLocation')
+    * call read('classpath:domain/edge-rtac/features/util/initData.feature@PostHoldings')
+    * def itemStatusName = 'Checked out'
+    * def createSecondItemResponse = call read('classpath:domain/edge-rtac/features/util/initData.feature@PostItem') { itemId: 68640328-788e-53fd-9c3c-af39e243f3c7 }
+    * def periodicalStatus = true
+    * def rtacBatchrequest = read('samples/rtac-batch-request-entity.json')
+
+    Given path 'rtac-batch'
```
marcjohnson-kint, line 94: The scenario name states `WHEN &fullPeriodicals=true`, is that intended to be a parameter to the request to edge-rtac?

edge-rtac/src/main/resources/domain/edge-rtac/rtac-junit.feature
```diff
@@ -0,0 +1,20 @@
...
+      | 'mod-permissions'         |
+      | 'edge-rtac'               |
+      | 'mod-inventory'           |
+      | 'mod-inventory-storage'   |
+
+    * table adminAdditionalPermissions
+      | name                      |
+
+    * table userPermissions
+      | name                      |
```
marcjohnson-kint, line 17: These tests create instances, holdings, items, locations etc, how does it do that when there are no permissions granted to the users?

edge-rtac/src/main/resources/domain/edge-rtac/features/util/initData.feature
```diff
@@ -0,0 +1,122 @@
+Feature: init data for mod-circulation
```
marcjohnson-kint, line 1: Should this say edge-rtac instead of mod-circulation

```suggestion
Feature: init data for edge-rtac
```

edge-rtac/src/main/resources/domain/edge-rtac/rtac-junit.feature
SinghAdes: with testAdmin there is no need to add any permissions, whatever modules are added in *** table modules** by default all permissions will get added automatically to the user. 
Let me know if thats okay or we should use **testUser** only.

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
SinghAdes: i guess no, **fullPeriodicals=true** is added in  request payload file. PFB 

  "instanceIds": [
    "#(firstInstanceId)",
    "#(secondInstanceId)"
  ],
  **"fullPeriodicals": "#(periodicalStatus)"**

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
SinghAdes: api response is get shuffled for each test run. 
And i think test is just for checking weather its returning status & detail or not.

edge-rtac/src/main/resources/domain/edge-rtac/rtac-junit.feature
marcjohnson-kint: > with testAdmin there is no need to add any permissions, whatever modules are added in *** table modules** by default all permissions will get added automatically to the user.

Oh, I didn't know that was done automatically.

> Let me know if thats okay or we should use testUser only.

Given we have the separation. I would expect that all of the preparation e.g. creating items, fees / fines etc to be done by the administrative user and the use of the API being tested by the regular user.

Is that how these tests work at the moment?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
marcjohnson-kint: The edge-rtac API includes this [parameter](https://github.com/folio-org/edge-rtac/blob/46e58fcc18be555c5ec84658012d1367eb8d77a0/ramls/edge-rtac.raml#L29)

Have you asked the PO how these scenarios are intended to use the periodical information?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
marcjohnson-kint: > api response is get shuffled for each test run.

What do you mean by shuffled?

> And i think test is just for checking weather its returning status & detail or not.

I think the test should be specific that each holding has the expected status

PR #370, opened 2021-12-03
[MODORDERS-626] Unopening and changing fund distribution

damien-git:
## Purpose
[MODORDERS-626](https://issues.folio.org/browse/MODORDERS-626) - Can't reopen an order after changing the expense class

## Approach
Added a test matching the steps to reproduce the bug:
-  Create an order and an order line
-  Open the order
-  Unopen the order
-  Change the expense class
-  Open the order


damien-git:
Screenshot_20211206_093949

grabsefx:
image

PR #374, opened 2021-12-09
UXPROD-3394-Karate-Test-for-EDGE-RTAC

SinghAdes:
**If **mode of issuance** of an instance is **serial** type and item's material type is  newspaper or magazine or journal then it is considered as a periodical data.**
Periodical data is a newspaper or magazine. Periodical data typically has holdings information that describes the library's coverage. Example: https://newcatalog.library.cornell.edu/catalog/1545844.

When we send parameter fullPeriodicals = true for **periodical data** ,it returns instance & items information along with status.
When we send parameter fullPeriodicals = false or no parameter  for **periodical data**, it returns instance & holding information but not items information.


edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
```diff
@@ -0,0 +1,156 @@
...
+    And match response.instances.holdings.length == 2
+    And match [firstInstanceId,secondInstanceId] contains call expectedData response.instances.holdings,'instances'
+    And match [expectedFirstItemId,expectedSecondItemId] contains call expectedData response.instances.holdings,'holdings'
+    And match ['Available','Checked out'] contains call expectedData response.instances.holdings,'status'
+    #deleteItem
+    * call read('classpath:domain/edge-rtac/features/util/initData.feature@DeleteItems')
+
+  Scenario:For periodical/serial, return holdings and item information including availability for each instance UUID included in request WHEN &fullPeriodicals=true
+#   1st instance
+    * def modeOfIssuanceId = '068b5344-e2a6-40df-9186-1829e13cd344'
```
marcjohnson-kint, line 74: Is this what distinguishes a periodical from a non-periodical?

How does the implementation identify them, does it use hard-coded modes of issuance?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
```diff
@@ -0,0 +1,156 @@
...
+    And path 'rtac/' + instanceId
+    And param apikey = apikey
+    When method GET
+    Then status 200
+    And match response.holdings.holding.length == 2
+    # api response is getting shuffled for each run, matching response by converting into an array.
+    And match [expectedFirstItemId,expectedSecondItemId] contains call expectedData response.holdings,'holdings'
+    And match ['Available','Checked out'] contains call expectedData response.holdings,'status'
+    #delete created Items
+    * call read('classpath:domain/edge-rtac/features/util/initData.feature@DeleteItems')
```
marcjohnson-kint, line 34: Why is it necessary to delete the items within the test?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
```diff
@@ -0,0 +1,156 @@
...
+    Then status 200
+    And print response
+    And match response.instances.holdings.length == 2
+    And match [firstInstanceId,secondInstanceId] contains call expectedData response.instances.holdings,'instances'
+    And match [expectedFirstItemId,expectedSecondItemId] contains call expectedData response.instances.holdings,'holdings'
+    And match ['Available','Checked out'] contains call expectedData response.instances.holdings,'status'
+    #deleteItem
+    * call read('classpath:domain/edge-rtac/features/util/initData.feature@DeleteItems')
+
+  Scenario:For periodical/serial, return holdings and item information including availability for each instance UUID included in request WHEN &fullPeriodicals=true
```
marcjohnson-kint, line 72: Please format these consistently. Some have a space after the colon (which I prefer) others do not.

edge-rtac/src/main/resources/domain/edge-rtac/features/util/initData.feature
```diff
@@ -0,0 +1,174 @@
...
+    * url baseUrl
+    * callonce login { tenant: 'diku', name: 'diku_admin', password: 'admin' }
+    * def headers = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': 'application/json, text/plain' }
+
+
+  @PostModeOfIssuance
+  Scenario: create mode of issuance
+    * def modeOfIssuanceId = call random_uuid
+    * def instanceModeOfIssuanceRequest = read('samples/instance/issuance-mode-request-entity.json')
+    * instanceModeOfIssuanceRequest.name = instanceModeOfIssuanceRequest.name + ' ' + random_string()
```
marcjohnson-kint, line 13: Why does a random string need to be appended to the name?

edge-rtac/src/main/resources/domain/edge-rtac/rtac-junit.feature
```diff
@@ -0,0 +1,20 @@
...
+      | 'edge-rtac'               |
+
+    * table adminAdditionalPermissions
+      | name                      |
+
+    * table userPermissions
+      | name                      |
+
+  Scenario: create tenant and users for testing
+    * call login { tenant: 'diku', name: 'diku_admin', password: 'admin' }
```
marcjohnson-kint, line 20: Please add a comment describing why the diku (rather than tenant created for the Karate tests) is used

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
SinghAdes: Yes , it is necessary to delete because we can not create multiple materialtypes with same name in same tenant(Diku). 
So to create a new Materialtype for each run we have to delete previous materialtype and to delete materialtype it is necessary that it should not attached to any of the item.

edge-rtac/src/main/resources/domain/edge-rtac/features/util/initData.feature
SinghAdes: **Serial** type issuance we already have in Diku tenant so creating random mode of issuance to consider it as a non-serial mode.

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
SinghAdes: > Is this what distinguishes a periodical from a non-periodical?

Yes , as i am aware of . as per PO :- **Periodical data is a newspaper or magazine**, but going through mod-rtac code i found there is mode of issuance for Instance also which make difference between periodical & non-periodical data.

SinghAdes: > How does the implementation identify them, does it use hard-coded modes of issuance?

we can not create Serial mode of issuance because it  is already there in Diku tenent , thats why i have hardcoded it .

edge-rtac/src/main/resources/domain/edge-rtac/rtac-junit.feature
SinghAdes: Comment added.

edge-rtac/src/main/resources/domain/edge-rtac/features/util/initData.feature
marcjohnson-kint: > Serial type issuance we already have in Diku tenant so creating random mode of issuance to consider it as a non-serial mode.

Alternatively, use another of the existing modes of issuance that is appropriate for non-serial instances (ask the PO for which one) e.g. monograph.

Or, create a new mode of issuance with an appropriate name describing the intent (non-serial) rather than random.

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
marcjohnson-kint: > Yes , it is necessary to delete because we can not create multiple materialtypes with same name in same tenant(Diku).

Rather than always try to create these records. The tests could check if a record exists with the expected name and if it does, use that record instead of creating a new one.

This means that record creation and deletion can be avoided.

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
marcjohnson-kint: I don't understand this answer.

My question was intended to be: How does edge-rtac (or mod-rtac) determine whether something is a periodical / serial?

marcjohnson-kint: > Yes , as i am aware of . as per PO :- Periodical data is a newspaper or magazine, but going through mod-rtac code i found there is mode of issuance for Instance also which make difference between periodical & non-periodical data.

Thank you, that's what I was curious about.

Does that mean that the code in mod-rtac hard codes this ID as well?

SinghAdes: No , mod-rtac does not harcode ID but it hardcode value "Serial".

https://github.com/folio-org/mod-rtac/blob/master/src/main/java/org/folio/mappers/FolioToRtacMapper.java#:~:text=private%20boolean%20isPeriodical,%7D

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
SinghAdes: i dont think it is possible , Karate does not provide proper control statements like if-else. Only way to do it by using **EVAL** (javascript) which i dont think is good to put JS code in feature file.

edge-rtac/src/main/resources/domain/edge-rtac/features/util/initData.feature
SinghAdes: Yes I agree. Thankyou

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
marcjohnson-kint: I think that is a shame, having to create different records each time and delete them in each test adds a bunch of complexity to these tests.

If we can't do it, please add comments to all of the places where we create new records and delete them (and the records that reference them) to explain why the code is this way.

marcjohnson-kint: Is there anything in [this documentation](https://github.com/karatelabs/karate#conditional-logic) that could help?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
marcjohnson-kint: Thanks, that is useful to know.

Please add a comment in the tests that explain that it is the name of the mode of issuance that is important.

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
```diff
@@ -0,0 +1,153 @@
+Feature: rtac tests
+
```
K-Felk, line 2: It looks like all these tests are creating and attaching items.  Yet there is behavior specified in edge-rtac for returning holdings records if there are no attached item records (I know cos I put it in :).  Have I missed the test for this?  If not, one needs to be added.

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
SinghAdes: > Is there anything in [this documentation](https://github.com/karatelabs/karate#conditional-logic) that could help?

Karate provides **ternary if else** only but that will not be sufficient as we need to execute below statement 
** * callonce read('classpath:domain/edge-rtac/features/util/initData.feature@PostMaterialType')**  which is causing problem

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
SinghAdes: @K-Felk  
Yes , when we send param fullperiodicals=false then api only returns holdings data for periodical data , no matter if item is attached to the holdings or not.

EDGRTAC-48 :- **Create Karate test for Edge-RTAC: For periodical/serial, return only holdings information including availability for each instance UUID included in request WHEN &fullPeriodicals=false OR no parameter is omitted (edge-rtac-4)**

Is this you are talking about or i am mistaking ?

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
K-Felk: I think if you are doing one for newspaper material type you should also do one for journal.

K-Felk:
> **If **mode of issuance** of an instance is **serial** type and item's material type is newspaper or magazine or journal then it is considered as a periodical data.** Periodical data is a newspaper or magazine. Periodical data typically has holdings information that describes the library's coverage. Example: https://newcatalog.library.cornell.edu/catalog/1545844.

Clarification:  This is an OR check, not an AND.  If a record has ANY of the following, it is treated as a periodical by edge-rtac and mod-rtac:

1. A mode of issuance that equals "serial"
2. Material type that include "newspaper"
3. Material type that include "journal"

Having any one of those will cause the rtac modules to treat the record as a periodical.

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
K-Felk: You're right, I missed it.  Mea culpa.

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
K-Felk: FYI, I did a bit of research on this, and the standard response to people asking how to do branching and conditionals in Karate seems to be "you don't." ¯\_(ツ)_/¯

edge-rtac/src/main/resources/domain/edge-rtac/features/rtac.feature
SinghAdes: added "journal" materialType

K-Felk (approved):

PR #376, opened 2021-12-14
[MODINVOICE-346] Approve invoice with negative line

damien-git:
## Purpose
[MODINVOICE-346](https://issues.folio.org/browse/MODINVOICE-346) - Approving an invoice with multiple lines that ref the same POL 2 payment 1 credit. Causes over reported encumbrance value on budget

## Approach
Added a new test. It fails randomly before the fix.
Also fixed an existing issue with permissions (inventory.items.item.get).

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/cross-modules/cross-modules-junit.feature
+++ b/acquisitions/src/main/resources/thunderjet/cross-modules/cross-modules-junit.feature
@@ -18,13 +18,14 @@ Feature: cross-module integration tests
       | 'orders-storage.module.all'                                 |
 
     * table userPermissions
-      | name                                                        |
-      | 'invoice.all'                                               |
-      | 'orders.all'                                                |
-      | 'finance.all'                                               |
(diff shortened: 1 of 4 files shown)


acquisitions/src/main/resources/thunderjet/cross-modules/cross-modules-junit.feature
```diff
@@ -18,13 +18,14 @@ Feature: cross-module integration tests
...
-      | 'orders.item.reopen'  |
-      | 'orders.item.unopen'  |
+      | name                       |
+      | 'invoice.all'              |
+      | 'orders.all'               |
+      | 'finance.all'              |
+      | 'orders.item.approve'      |
+      | 'orders.item.reopen'       |
+      | 'orders.item.unopen'       |
+      | 'inventory.items.item.get' |
```
grabsefx, line 28: adding this permission here means that it was missed in module descriptor in mod-orders

grabsefx (approved):

PR #379, opened 2021-12-16
 Added "Create a "first available item" request via an external form"

SinghAdes:
Test:-
Create Karate test for Edge-Patron: Create a "first available item" request via an external form (edge-patron-7)


edge-patron/src/main/resources/domain/edge-patron/features/patrons.feature
```diff
@@ -25,7 +25,7 @@ Feature: patron tests
     * call read('classpath:domain/edge-patron/features/util/initData.feature@PostOwnerAndCharges')
 
     Given url edgeUrl
-    And path 'patron/account/' + extSystemId
+    And path 'patron/account/' + extSystemId + '?includeCharges=true'
```
marcjohnson-kint, line 28: Why has this request changed?

edge-patron/src/main/resources/domain/edge-patron/features/patrons.feature
```diff
@@ -94,3 +94,23 @@ Feature: patron tests
     Then status 200
     Then match response.totalHolds == 1
     And match response.holds[0].item.itemId == itemId
+
+   Scenario: Create a "first available item" request via an external form.
```
marcjohnson-kint, line 98: There is no form involved in the back end API, I suggest not including that in the description.

The important part of this scenario is that there is only a single item associated with the instance being requested.

```suggestion
   Scenario: Instance level request is associated with only item
```

edge-patron/src/main/resources/domain/edge-patron/features/patrons.feature
SinghAdes: That was a  mistake. its been changed now.

marcjohnson-kint (approved):

PR #394, opened 2022-01-03
Fix testing admin user so perms.users.assign.immutable FAT-1144

adamdickmeiss:


steveellis (approved):
If the new query returns the missing permission which then gets included in POST to perms/users then this is fine. If I'm understanding this is what is blocking MODLOGSAML-107. 

Maybe obvious, but I would only make the change in `setup-users.feature` if this new permission is something that all karate tests need. Otherwise each test project can request the permissions it needs via its `-junit.feature` file. For example [here's that file](https://github.com/folio-org/folio-integration-tests/blob/master/mod-user-import/src/main/resources/core_platform/mod-user-import/user-import-junit.feature) for mod-user import. These are what are merged in `setup-users.feature`.

Right now I'm unable to see this code work because no karate tests are working for me on `testing` nor `snapshot`. When I run `./runtests.sh mod-user-import snapshot` (for example) I get a bunch of `No suitable module found for path` errors in `setup-users.feature`. I could set up a local environment if you'd like someone else to verify it. Maybe I'm just forgetting something about how to run these tests?

adamdickmeiss:
> If the new query returns the missing permission which then gets included in POST to perms/users then this is fine. If I'm understanding this is what is blocking MODLOGSAML-107.
> 

> Maybe obvious, but I would only make the change in `setup-users.feature` if this new permission is something that all karate tests need. Otherwise each test project can request the permissions it needs via its `-junit.feature` file. For example [here's that file](https://github.com/folio-org/folio-integration-tests/blob/master/mod-user-import/src/main/resources/core_platform/mod-user-import/user-import-junit.feature) for mod-user import. These are what are merged in `setup-users.feature`.

I have not verified, but I think many modules uses this facility to make a test user with permissions that is suitable for testing their module.. Without perms.users.assign.immutable that will fail. There's also a related 
https://issues.folio.org/browse/MODPATRON-98 . (there could be others that I haven't found yet).


> 
> Right now I'm unable to see this code work because no karate tests are working for me on `testing` nor `snapshot`. When I run `./runtests.sh mod-user-import snapshot` (for example) I get a bunch of `No suitable module found for path` errors in `setup-users.feature`. I could set up a local environment if you'd like someone else to verify it. Maybe I'm just forgetting something about how to run these tests?

Yes.. Normally one would run against, snapshot, Vagrant box, Rancher.. or other.. In my case, because I wanted to debug https://issues.folio.org/browse/MODLOGSAML-107 I created a script for this.. https://issues.folio.org/secure/attachment/43556/modlogsaml-107.sh

common/src/main/resources/common/setup-users.feature
```diff
@@ -78,7 +78,7 @@ Feature: prepare data for api test
     Given path '/perms/permissions'
     And header x-okapi-tenant = testTenant
     And param length = 1000
-    And param query = '(subPermissions="" NOT subPermissions ==/respectAccents []) and (cql.allRecords=1 NOT childOf <>/respectAccents [])'
+    And param query = 'subPermissions="" and (cql.allRecords=1 NOT childOf <>/respectAccents [])'
```
julianladisch, line 81:
```suggestion
    And param query = 'childOf == "[]"'
```

RMB always creates an emtpy array if the client doesn't provide a `childOf` property. Therefore we don't need the `NOT` construct to also match a missing `childOf` property.

`subPermissions=""` is the same as `cql.allRecords=1` because RMB always creates an empty array if the client doesn't provide a `subPermissions` property.

`/respectAccents` is valid CQL syntax but ignored by RMB. To influence the matching it must be specified in schema.json.

The JSON array `[]` brackets are always without accents, therefore `/respectAccents` is not needed.

PR #404, opened 2022-01-07
MODUSERS-292 search for a user by username in the keyword box 

SinghAdes:


marcjohnson-kint (commented):
Thanks for the quick work on this.

(It's a bit of a pity that the URL is encoded behind the scenes. I understand that is a limitation of the tool and I appreciate the comment explaining this)

mod-users/src/main/resources/domain/mod-users/features/users.feature
```diff
@@ -142,6 +142,22 @@ Feature: Users tests
     And match response.users[0].username == 'aaa'
     And match response.totalRecords == 1
 
+    Scenario: Search user by username in the keyword box.
```
marcjohnson-kint, line 145:
```suggestion
    Scenario: Use keyword search to find a user by username.
```

marcjohnson-kint (approved):

PR #405, opened 2022-01-10
Mod user bl karate tests permission issue

SinghAdes:
@adamdickmeiss 
It would seem that mod-users-bl cannot be enabled before mod-authtoken, otherwise permissions problems will happen.
So i have added 1 extra feature file (configurepermission.feature) to set permissions for admin. 

The permission issue is occurring at test:- Scenario: Return a composite object for the currently logged in user by username(Line no:-97) in users-bl.feature file.


adamdickmeiss:
This PR should be closed as it's been superseded with #408

PR #410, opened 2022-01-12
[MODORDERS-638] Pay invoice with new expense class

damien-git:
## Purpose
[MODORDERS-638](https://issues.folio.org/browse/MODORDERS-638) - Invoice can be approved but not paid when expense class added at point of invoice

## Approach
Added a new cross-module test following the scenario in MODORDERS-638. Before the MODORDERS-638 change, it fails when the order line is updated. In older versions of FOLIO it failed when the invoice was paid.
Also removed a user permission that is no longer necessary.


AndreiMPV:
image

PR #415, opened 2022-01-17
mod-users: fix user already exists

adamdickmeiss:
https://issues.folio.org/browse/MODUSERS-300


SinghAdes (commented):
i have made changes to this scenario, kindly review.
https://github.com/folio-org/folio-integration-tests/pull/416

PR #427, opened 2022-01-24
[MODINVOICE-270] Cancel invoice

damien-git:
## Purpose
[MODINVOICE-270](https://issues.folio.org/browse/MODINVOICE-270) - Cancel invoice : create logic to support invoice cancel operation

## Approach
- Added a new test to cancel an invoice and check results.
- Fixed 2 unrelated integration tests, check-invoice-and-invoice-lines-deletion-restrictions and check-that-not-possible-pay-for-invoice-without-approved.
- Added transaction and budget checks for [MODFISTO-273](https://issues.folio.org/browse/MODFISTO-273).


AndreiMPV:
Verified on testing
image

PR #429, opened 2022-01-26
[MODFISTO-273] - Cansel invoice cross-modules test

Dmitriy-Butramyou:


damien-git:
These tests were merged with the [MODINVOICE-270 PR](https://github.com/folio-org/folio-integration-tests/pull/427).

PR #448, opened 2022-02-14
[MODORDERS-645] Remove and add order line fund distribution before invoice approval

damien-git:
## Purpose
[MODORDERS-645](https://issues.folio.org/browse/MODORDERS-645) - Invoice payment fails if order line fund distribution is changed before invoice approval

## Approach
Added a new cross-module test.


AndreiMPV:
image

PR #470, opened 2022-03-10
Fat 943

SatishkumarPurswani:


Abdulkhakimov:
I think would be better to display request body in scenario instead of moving to json files. It is very difficult to go every time to read the body

SatishkumarPurswani:
@Abdulkhakimov - 
1. Json files moved to separate files are larger in size, it may affect the readability, thinking the same I had to move it to separate file.
2. ActionProfile JSON is common between the scenarios, so I had to move it to a separate file.
I could move JSON's mentioned in 1st point to the feature file, Please confirm.TIA.

Abdulkhakimov:
> @Abdulkhakimov -
> 
> 1. Json files moved to separate files are larger in size, it may affect the readability, thinking the same I had to move it to separate file.
> 2. ActionProfile JSON is common between the scenarios, so I had to move it to a separate file.
>    I could move JSON's mentioned in 1st point to the feature file, Please confirm.TIA.

Hi @SatishkumarPurswani. We discussed with @VRohach and decided not to move the request body into separate json files. The scenarios are the same but the profiles are different. In addition, it will be difficult for us to find the right request every time. Thank you.

SatishkumarPurswani:
> > @Abdulkhakimov -
> > 
> > 1. Json files moved to separate files are larger in size, it may affect the readability, thinking the same I had to move it to separate file.
> > 2. ActionProfile JSON is common between the scenarios, so I had to move it to a separate file.
> >    I could move JSON's mentioned in 1st point to the feature file, Please confirm.TIA.
> 
> Hi @SatishkumarPurswani. We discussed with @VRohach and decided not to move the request body into separate json files. The scenarios are the same but the profiles are different. In addition, it will be difficult for us to find the right request every time. Thank you.

Hi @Abdulkhakimov - I have done the changes as per our discussion. Request you to please review.

TetianaParanich (approved):

PR #484, opened 2022-03-23
FAT-1517 Karate Tests: Create MARC authority records via data import

YahorSviarchkou:
### Learning

[FAT-1517 Create MARC authority records via data import](https://issues.folio.org/browse/FAT-1517)


ALiemeshka (approved):

data-import/src/main/resources/folijet/data-import/features/create-records.feature
```diff
@@ -0,0 +1,50 @@
...
+    Given call read(utilFeature+'@ImportRecord') { fileName:'marcAuthorityInvalid1XX', jobName:'createAuthority' }
+    Then match status != 'ERROR'
+
+    Given path '/source-storage/source-records'
+    And param recordType = 'MARC_AUTHORITY'
+    And param snapshotId = jobExecutionId
+    And headers headersUser
+    And retry until response.totalRecords > 0 && karate.sizeOf(response.sourceRecords[0].externalIdsHolder) > 0
+    When method get
+    Then status 200
```
psmagin, line 41: If data-import is failed then why do you check source-storage contains some record. You should check that record doesn't exist

data-import/src/main/resources/folijet/data-import/features/create-records.feature
```diff
@@ -0,0 +1,50 @@
+Feature: Create records
```
psmagin, line 1:
```suggestion
Feature: Create marc authority records
```

psmagin: and the save for file name

psmagin (changes requested):

data-import/src/main/resources/folijet/data-import/features/create-records.feature
YahorSviarchkou: > If data-import is failed then why do you check source-storage contains some record. You should check that record doesn't exist

Date import can create authority record, even if the record contains not valid 1xx

psmagin (approved):

PR #488, opened 2022-03-25
MODORDERS-660 Improve validation to preventing creation order lines with wrong expense classes

a-churakov:
https://issues.folio.org/browse/MODORDERS-660


AndreiMPV:
Tested on snapshot
image
image

PR #505, opened 2022-04-04
Rancher 239

VitalyDemchenkoEpam:
Add teams assignment json
Disable report for "runFeature" used in @Before and @After to avoid redundant reports in Jenkins


marcjohnson-kint:
@VitalyDemchenkoEpam 

I thought it had been decided a while ago that teams would be decided by the name of the team in the path for the tests. Does this change supersede / override that decision? And now team / module assignments will be decided by configuration?

testrail-integration/src/main/java/org/folio/test/TestBase.java
```diff
@@ -13,79 +13,86 @@
...
-      testFeatureName = testFeatureName.concat(".feature");
+
+    @BeforeAll
+    public void beforeAll() {
+        runHook();
+    }
+
+    @AfterAll
+    public void afterAll() {
+    }
```
marcjohnson-kint, line 89: Why is an empty after all step needed?

teams-assignment.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  {
+    "team": "core_platform",
+    "modules": ["mod-configuration", "mod-permissions", "mod-login-saml", "mod-user-import"],
+    "slackChannel": "#core-platform"
+  },
+  {
+    "team": "falcon",
+    "modules": ["mod-search"],
+    "slackChannel": "#falcon"
```
psmagin, line 15: Should be in spitfire team

PR #521, opened 2022-04-11
[MODFISTO-293] Approve invoice using different fiscal years

damien-git:
## Purpose
[MODINVOICE-387](https://issues.folio.org/browse/MODINVOICE-387) - "Insufficient fund validation to approve invoice" and [MODFISTO-293](https://issues.folio.org/browse/MODFISTO-293) - "Error All expected transactions already processed"

## Approach
Create an invoice that will not approve because it is using 2 different fiscal years. The insufficient validation caused by MODINVOICE-387 will result in mod-finance-storage being called with 3 good transactions and 1 bad transaction. Then, without the MODFISTO-293 fix, mod-finance-storage will fail to process the transactions again after the invoice has been fixed.

Note that some randomness is involved, so it is possible that the test will pass even without the MODFISTO-293 fix. Using 4 lines and trying to approve the invoice several times increases the chances of failure.


AndreiMPV:
Verified on snapshot 
image

AndreiMPV (approved):

PR #537, opened 2022-04-20
FAT-1508: Ensure request queue containing item and title requests is ordered correctly

yusuf-murodov-epam:
## Purpose
[FAT-1508](https://issues.folio.org/browse/FAT-1508)
_Ensure request queue containing item and title requests is ordered correctly_

## Approach


mod-circulation/src/main/resources/vega/mod-circulation/features/requests.feature
```diff
@@ -702,4 +702,56 @@ Feature: Requests tests
...
+    * call read('classpath:vega/mod-circulation/features/util/initData.feature@PostHoldings') { extHoldingsRecordId: #(extHoldingId), extInstanceId: #(extInstanceId) }
+
+    # post the first item and a page tlr
+    * call read('classpath:vega/mod-circulation/features/util/initData.feature@PostItem') { extItemId: #(extItemId1), extItemBarcode: #('FAT-1508IBC-1'), extHoldingsRecordId: #(extHoldingId) }
+    * def extRequestId1 = call uuid1
+    * def tlrResponse = call read('classpath:vega/mod-circulation/features/util/initData.feature@PostTitleLevelRequest') { requestId: #(extRequestId1), requesterId: #(extUserId1), extInstanceId: #(extInstanceId) }
+
+    # post the second item, a page ilr and a hold tlr
+    * call read('classpath:vega/mod-circulation/features/util/initData.feature@PostItem') { extItemId: #(extItemId2), extItemBarcode: #('FAT-1508IBC-2'), extHoldingsRecordId: #(extHoldingId) }
+    * def extRequestId2 = call uuid1
```
Mykyta-Varenyk, line 737: How about creating first hold tlr and then page ilr, so that no matter the order of creation the queue works as needed?

yusuf-murodov-epam: there is not allowed to create hold tlr if there is available item in the instance, so I created page ilr for the second item, then created hold tlr for the instance

Mykyta-Varenyk: Exactly, ok

Mykyta-Varenyk (approved):

PR #549, opened 2022-04-28
[MODFISTO-298] Partial rollover

damien-git:
## Purpose
[MODFISTO-298](https://issues.folio.org/browse/MODFISTO-298) - Unable to access adjustment ledger

## Approach
- New test of a rollover with no order type selected for encumbrances
- New reusable features
- Fixed a permission issue


AndreiMPV:
On snapshot
image

PR #590, opened 2022-05-31
FAT-1024: When patron has exceeded their Patron Group Limit for 'Recall overdue by maximum number of days', patron is not allowed to borrow items per Conditions settings

yusuf-murodov-epam:
## Purpose
[FAT-1024](https://issues.folio.org/browse/FAT-1024)
_When patron has exceeded their Patron Group Limit for 'Recall overdue by maximum number of days', patron is not allowed to borrow items per Conditions settings_

## Approach


mod-circulation/src/main/resources/vega/mod-circulation/features/loans.feature
```diff
@@ -1307,3 +1307,90 @@ Feature: Loans tests
...
+    * call read('classpath:vega/mod-circulation/features/util/initData.feature@PostGroup') { extUserGroupId: #(groupId) }
+    * call read('classpath:vega/mod-circulation/features/util/initData.feature@PostUser') { extUserId: #(extUserId1), extUserBarcode: #(extUserBarcode1), extGroupId: #(groupId) }
+    * call read('classpath:vega/mod-circulation/features/util/initData.feature@PostUser') { extUserId: #(extUserId2), extUserBarcode: #(extUserBarcode2), extGroupId: #(groupId) }
+
+    # get the interested condition id, Recall overdue by maximum number of days
+    Given path 'patron-block-conditions'
+    When method GET
+    Then status 200
+    * def conditions = karate.sort(response.patronBlockConditions, (condition) => condition.name)
+    * def conditionId = conditions[5].id
```
Mykyta-Varenyk, line 1339: What if someday in the future we are going to have a new condition type? Maybe we should fine the needed condition in some other way?

yusuf-murodov-epam: fixed

Mykyta-Varenyk (approved):

PR #612, opened 2022-06-13
[MODORDERS-712] Rollover with closed order

damien-git:
## Purpose
[MODORDERS-712](https://issues.folio.org/browse/MODORDERS-712) - Encumbrances are created for closed orders during FYRO

## Approach
- Added a new test with a rollover using a closed order
- Used a new reusable feature to close an order, but this is already added by the integration test for MODORDERS-706


Dmitriy-Butramyou:
image

Dmitriy-Butramyou (approved):

PR #621, opened 2022-06-17
[FAT-1540] Implement API Karate tests - MARC transformation options

aleksandr2110:
Purpose

Karate implementation is created for the following endpoints

GET /central-servers/{centralServerId}/marc-transformation-options
POST /central-servers/{centralServerId}/marc-transformation-options
PUT /central-servers/{centralServerId}/marc-transformation-options
DELETE /central-servers/{centralServerId}/marc-transformation-options

PR #622, opened 2022-06-20
Revert "[FAT-1540] Implement API Karate tests - MARC transformation options"

aleksandr2110:
Reverts folio-org/folio-integration-tests#621


aleksandr2110:
With error

PR #653, opened 2022-07-11
FAT-2102 - Karate test fail: [folijet/data-import] DataImportApiTest {2}  folijet/data-import/features/data-import-integration.feature

RuslanLavrov:
## Approach
* FAT-937 - fix instance.catalogedDate mapping
* FAT-944 FAT-945:
  * extracted steps for FAT-937 to separate .feature file and reused from FAT-944, FAT-945 for data preparation
  * corrected scenarios to export imported marc record and to use exported record for further import in the tests
  * added data-export mapping profile template to export holding and item hrid as well
  * fixed match profile for holding
  * removed unused .csv files with hardcoded instance id

## Learning
[FAT-2102](https://issues.folio.org/browse/FAT-2102)

image


KaterynaSenchenko (approved):

data-import/src/main/resources/folijet/data-import/global/import-instance-holding-item.feature
```diff
@@ -0,0 +1,544 @@
+Feature: Util feature to import instance, holding, item. Based on FAT-937 scenario steps.
```
VRohach, line 1: @RuslanLavrov  Let's rename this file such as "deafult-import.-instance-holding-item.feature". It is needed because there are a lot of selected values inside mapping profiles,  which can be changed in other cases.

VRohach (approved):

Aliaksandr-Fedasiuk (approved):

data-import/src/main/resources/folijet/data-import/global/import-instance-holding-item.feature
RuslanLavrov: done

PR #654, opened 2022-07-11
MODORDERS-619 added retrieving titles test

grabsefx:
image


acquisitions/src/main/resources/thunderjet/mod-orders/orders.feature
```diff
@@ -175,5 +177,8 @@ Feature: mod-orders integration tests
   Scenario: Validate fund distribution for zero price
     Given call read("features/productIds-field-error-when-attempting-to-update-unmodified-order.feature")
 
+  Scenario: Validate fund distribution for zero price
```
SerhiiNosko, line 180: I think this scenario should be renamed

SerhiiNosko (approved):

PR #687, opened 2022-09-04
FAT-2320 Fix "Package Name already exists" error

maryna-steshenko:
## Purpose
Fix "Package Name already exists" and other similar errors that occur during execution of setup-resources feature for mod-kb-ebsco-java module tests.
## Approach
Apparently, an error occurs on EBSCO API side and since we cannot fix it directly a workaround was implemented.
- Different single package, package and title names are generated for each test execution
- Package test fixed

##Results:
image


oliinyko (approved):

mod-kb-ebsco-java/src/main/resources/spitfire/mod-kb-ebsco-java/features/packages.feature
```diff
@@ -118,7 +118,7 @@ Feature: Packages
     When method GET
     Then status 400
 
-  Scenario: POST Packages should return 400 if Package with the provided name already exists
+  Scenario: POST Packages should return 500 if Package with the provided name already exists
```
Shans-Kaluhin, line 121: I don't think we should expect server error when package name already exists

psmagin (approved):

PR #691, opened 2022-09-08
[FAT-2412] - Problem when waiting pending status of batch export

Abdulkhakimov:
## Purpose
The reason for the problem is that there was no resend request when the status of "Pending" jobs was received. Problem was occured only on karate autorun env because env has not enough resoursec.

## Aproach

- Retry responses to check batch-voucher-exports endpoint until get 'Error' status.

image


SerhiiNosko (approved):

acquisitions/src/main/resources/thunderjet/mod-invoice/features/check-vendor-address-included-with-batch-voucher.feature
```diff
@@ -106,11 +108,11 @@ Feature: Check vendor address included with batch voucher
     When method POST
     Then status 201
     * def batchVoucherExportId = $.id
-    * call pause 1000
 
     # ============= get export later to give it time to create the batch voucher ===================
     Given path 'batch-voucher/batch-voucher-exports', batchVoucherExportId
     And headers headersUser
+    And retry until response.status == 'Error'
```
SerhiiNosko, line 115: Should it be Completed status here?

Abdulkhakimov: I thought the same but I checked on the snapshot environment and got the error status

SerhiiNosko (approved):

Dmitriy-Butramyou (approved):

PR #695, opened 2022-09-13
[FAT-2408] -  Missed param when call createFund function

Abdulkhakimov:
## Purpose
When we call createFund feature in the "current budget for fund" main scenario we should specify "code" param if not code will be retrieved from response.

## Learning
[FAT-2408](https://issues.folio.org/browse/FAT-2408)


Abdulkhakimov:
https://issues.folio.org/browse/FAT-2406

Dmitriy-Butramyou (approved):

PR #705, opened 2022-09-19
FAT-2962: Fix invalid tenant ids

julianladisch:
It was decided to restrict the tenant id:

https://wiki.folio.org/display/TC/ADR-000002+-+Tenant+Id+and+Module+Name+Restrictions

    no underscore
    no upper case letters
    maximum length 31 characters
    regex: [a-z][a-z0-9]{0,30}

Some tenants in folio-integration-tests are no longer allowed.


adamdickmeiss (approved):
LGTM. Time will tell if it suffice.

PR #714, opened 2022-09-22
Karate tests for age-to-lost items are commented

bekhzodkurbonboev:
Tests related to aged to lost loans are commented: they are taking more time than expected.

JIRA: https://issues.folio.org/browse/FAT-2145


mod-circulation/src/main/resources/vega/mod-circulation/features/loans.feature
```diff
@@ -779,52 +779,52 @@ Feature: Loans tests
...
-
-    # revert retry configuration to default values
-    * configure retry = { count: 3, interval: 3000 }
-
-    # revert age-to-lost processor delay time
-    Given path '/_/proxy/tenants/' + tenant + '/timers'
-    And request '{"id":"mod-circulation_6","routingEntry":{"unit": "minute", "delay":"30"} }'
-    When method PATCH
-    Then status 204
+#  Scenario: When an existing loan is aged to lost update agedToLostDate, item status to Aged to lost
```
roman-barannyk, line 782: I think we  can use @ignore annotation, please check if it works

roman-barannyk (approved):

mod-circulation/src/main/resources/vega/mod-circulation/features/loans.feature
bekhzodkurbonboev: In other modules they use @ignore on Feature and @Ignore on Scenario, but when I use them Scenarios were not skipped.

PR #728, opened 2022-10-05
FAT-1645 - Bulk-edit(Users) move tests run to test tenant

obozhko-folio:
[FAT-1645](https://issues.folio.org/browse/FAT-1645) - Bulk-edit(Users) move tests run to test tenant

## Purpose
Use test tenant, fix karate tests for bulk-edit users.

## Approach
Add pauses, add permissions, update sample files, refactor code.

success


khandramai (approved):

mod-bulk-edit/src/main/resources/global/add-okapi-permissions.feature
```diff
@@ -0,0 +1,83 @@
+Feature: prepare data for api test
+
```
alekGbuz, line 2: Is it possible to use table adminAdditionalPermissions as common approach to add needed permissions for admin user,  but not own separate feature.

siarhei-charniak (approved):

alekGbuz (approved):

PR #746, opened 2022-10-18
FAT-3206: Upgrade vulnerable dependencies

julianladisch:
Upgrade cucumber-reporting from 5.7.1 to 5.7.3, this indirectly upgrades jsoup from 1.15.1 to 1.15.3 fixing cross-site scripting (XSS) https://nvd.nist.gov/vuln/detail/CVE-2022-36033

Upgrade graal-sdk from 21.2.0 to 21.3.3.1 fixing 18 vulnerabilities https://security.snyk.io/package/maven/org.graalvm.sdk:graal-sdk/21.2.0

Upgrade commons-codec from 1.11 to 1.15 fixing Information Exposure https://app.snyk.io/vuln/SNYK-JAVA-COMMONSCODEC-561518

Upgrade snakeyaml from 1.29 to 1.33 fixing Denial of Service (DoS) https://nvd.nist.gov/vuln/detail/CVE-2022-25857
https://nvd.nist.gov/vuln/detail/CVE-2022-38749

Upgrade jackson-databind from 2.13.0 to 2.13.4.2 fixing Denial of Service (DoS) https://nvd.nist.gov/vuln/detail/CVE-2022-42003

Upgrade Vert.x from unsupported version 4.2.7 to supported version 4.3.4.

PR #748, opened 2022-10-21
FAT-3213: Fix graal upgrade

julianladisch:
https://issues.folio.org/browse/FAT-3206 = #746 upgraded graal-sdk from 21.2.0 to 21.3.3.1 fixing 18 vulnerabilities.

It was forgotten to also upgrade org.graalvm.js:js-scriptengine and org.graalvm.js:js from 21.2.0 to 21.3.3.1. This makes folio-integration-test hang.

Fix: Also upgrade graal's js-scriptengine and js.


Shans-Kaluhin (approved):
Works good

PR #764, opened 2022-11-01
FAT-2132: Karate test fail: [prokopovych/mod-inventory]

julianladisch:
User proper param syntax in users.feature.
Move uuid1 function to common/util/*.feature.
Use util1 as function name space in utils.feature to avoid hiding the same function in inventoryFeatureTest.feature.


adamdickmeiss (approved):
I trust you (don't know the Karate tests stuff, much).

PR #783, opened 2022-11-14
mod-entities-links: linking rules feature

Shans-Kaluhin:
## Approach
- Move setup logic to utill feature
- Add test for linking-rules


mod-entities-links/src/main/resources/spitfire/mod-entities-links/features/linkingRulesTest.feature
```diff
@@ -0,0 +1,15 @@
...
+    * configure headers = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': 'application/json'  }
+    * def samplePath = 'classpath:spitfire/mod-entities-links/features/samples'
+
+  @Positive
+  Scenario: Get instance to authority rules - Should match json rules
+    * def jsonRules = read(samplePath + 'linking-rules.instance-authority.json')
+    Given path '/linking-rules/instance-authority'
+    When method GET
+    Then status 200
+    And match response == jsonRules
```
psmagin, line 15: I suggest not matching the full set of rules but just checking that rules exist in response. For full response check we will be required to update Karate tests with each change to linking rules

Shans-Kaluhin: Why not)?

psmagin (approved):

PR #799, opened 2022-11-28
FAT-3430: inventoryFeatureTest add locations

julianladisch:
Locations are no longer reference data, we need to create them in the inventoryFeatureTest.
https://issues.folio.org/browse/MODINVSTOR-443


adamdickmeiss (approved):
Why not move this to sample data, then. Now "sample data" is outside the repo (mod-inventory!).

PR #802, opened 2022-11-30
Fat 992: When patron and item id's entered at checkout, execute circulation rules and return the policy to be applied, matching the criteria with the highest priority or the fallback policy

bekhzodkurbonboev:
Implement Karate test for Scenario: "When patron and item id's entered at checkout, execute circulation rules and return the policy to be applied, matching the criteria with the highest priority or the fallback policy"

JIRA: https://issues.folio.org/browse/FAT-992


mod-circulation/src/main/resources/vega/mod-circulation/features/loans.feature
```diff
@@ -1988,4 +1988,60 @@ Feature: Loans tests
...
+    And params queryParams
+    When method GET
+    Then status 200
+    * def overdueFinePolicyToBeApplied = response.overdueFinePolicyId
+
+    Given path 'circulation', 'rules', 'lost-item-policy'
+    And params queryParams
+    When method GET
+    Then status 200
+    * def lostItemPolicyToBeApplied = response.lostItemPolicyId
```
roman-barannyk, line 2035: I think it makes sense to verify a notice policy and request policy as well

roman-barannyk (approved):

PR #803, opened 2022-11-30
FAT-3427: ModBulkEditApiTest add sample locations

julianladisch:
mod-inventory-storage has moved locations and location-units from reference data to sample data: https://issues.folio.org/browse/MODINVSTOR-443
https://github.com/folio-org/mod-inventory-storage/pull/850/files

We need to create all needed locations in mod_item_init_data.feature/mod-item-util.feature.


adamdickmeiss (approved):
So this data is in sample as well. And it doesn't bother you that Karate tests should just load sample?

This is a violation of DRY. Just saying :-)

PR #810, opened 2022-12-12
FAT-991: When a new circulation rule is entered in the circulation ed…

bekhzodkurbonboev:
Implemented Karate test for Scenario: "When a new circulation rule is entered in the circulation editor, add the rule to the circulation rules record"

JIRA: https://issues.folio.org/browse/FAT-991


mod-circulation/src/main/resources/vega/mod-circulation/features/loans.feature
```diff
@@ -2333,4 +2333,86 @@ Feature: Loans tests
...
+    Given path 'loan-policy-storage', 'loan-policies', newLoanPolicyId
+    When method GET
+    Then status 200
+
+    # enter new circulation rule in the circulation editor
+    * def rules = 'priority: t, s, c, b, a, m, g\nfallback-policy: l '+newLoanPolicyId+' r '+newRequestPolicyId+' n '+newNoticePolicyId+' o '+newOverdueFinePolicyId+' i '+newLostItemFeePolicyId
+    * def rulesEntityRequest = { "rulesAsText" : "#(rules)" }
+    Given path 'circulation', 'rules'
+    And request rulesEntityRequest
+    When method PUT
```
bekhzodkurbonboev, line 2403: When I try to PUT the "rules" It is returning an error. 
But If I PUT this "rules" on ' circulation-rules-storage' endpoint, then policies are being applied correctly so I think there is no error in formatting.
error message

alexanderkurash (approved):

PR #821, opened 2022-12-21
[MODORDERS-807] - Test for copy number

Dmitriy-Butramyou:
## Purpose

## Approach
The tests have been updated as the logic of creating holdings has been updated. Now if the piece contains only the 'LocationID' field when it is created, the corresponding 'holdingId' will be searched, if it is not, a new one will be created.

### TODOS and Open Questions

### Learning

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/features/update_copy_number.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/features/update_copy_number.feature
@@ -1,0 +1,104 @@
+@parallel=false
+# for https://issues.folio.org/browse/MODORDERS-807
+Feature: Should update copy number in item after updating in piece
+
+  Background:
+    * print karate.info.scenarioName
+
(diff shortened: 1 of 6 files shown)


acquisitions/src/main/resources/thunderjet/mod-orders/features/update_copy_number.feature
```diff
@@ -0,0 +1,104 @@
...
+    Then status 200
+    * def piece = $.pieces[0]
+    
+    Given path '/inventory/items/', piece.itemId
+    When method GET
+    Then status 200
+    * def item = $
+
+    And match piece.copyNumber == '12345'
+    And match item.copyNumber == '12345'
```
SerhiiNosko, line 104: Can you extend this test please to check not only copy number but other fields that we are updaing in mod-orders

SerhiiNosko (approved):

PR #850, opened 2023-01-31
[MODFIN-299] Group and ledger transfers after rollover

damien-git:
## Purpose
[MODFIN-299](https://issues.folio.org/browse/MODFIN-299) - Group and Ledger summary for Net Transfers do not include Rollover transfer amounts.

## Approach
Added a new test checking net transfers for group summaries and ledgers after a rollover using `rolloverAvailable`.

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-finance/features/group-and-ledger-transfers-after-rollover.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-finance/features/group-and-ledger-transfers-after-rollover.feature
@@ -1,0 +1,221 @@
+# For https://issues.folio.org/browse/MODFIN-299
+@parallel=false
+Feature: Group and ledger transfers after rollover
+
+  Background:
+    * print karate.info.scenarioName
+
+    * url baseUrl
+
+    * callonce login testUser
+    * def okapitokenUser = okapitoken
+
+    * def headersUser = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenUser)', 'Accept': 'application/json' }
+    * configure headers = headersUser
+
+    * callonce variables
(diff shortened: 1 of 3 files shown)


Abdulkhakimov (approved):

acquisitions/src/main/resources/thunderjet/mod-finance/features/group-and-ledger-transfers-after-rollover.feature
```diff
@@ -0,0 +1,221 @@
...
+    * def f1Y1BudgetId = callonce uuid8
+    * def f1Y2BudgetId = callonce uuid9
+    * def f2Y1BudgetId = callonce uuid10
+    * def f2Y2BudgetId = callonce uuid11
+    * def f3Y1BudgetId = callonce uuid12
+    * def f3Y2BudgetId = callonce uuid13
+    * def groupId = callonce uuid14
+    * def rolloverId = callonce uuid15
+
+    * def createFiscalYear = read('classpath:thunderjet/mod-finance/reusable/createFiscalYear.feature')
```
SerhiiNosko, line 35: I see that you not declaring functions like createFund, createLedger, createBudget , declaring only  createFiscalYear .
So this is optional step?

damien-git: These are already defined by [karate-config.js](https://github.com/folio-org/folio-integration-tests/blob/master/acquisitions/src/main/resources/karate-config.js). Eventually we might want to add more there to lighten the features, or remove them to make all dependencies explicit and to be more consistent. I am not sure how it would affect performance, but this should be taken into account.

SerhiiNosko (approved):

PR #852, opened 2023-02-03
[MODINVOICE-449]-Adjustments cause blocker on invoices.

SinghAdes:
## Purpose
https://issues.folio.org/browse/MODINVOICE-449
## Approach

### TODOS and Open Questions

### Learning


SinghAdes:
image

acquisitions/src/main/resources/thunderjet/mod-invoice/features/check-invoice-line-validation-with-adjustments.feature
```diff
@@ -0,0 +1,71 @@
+Feature: Check invoiceLine validation with  adjustments
```
SerhiiNosko, line 1: Please add jira story name, and I think this tests should be renamed to something like this: check-invoice-validation-for-adjustments-not-fails-with-correct-adjustments

SerhiiNosko: or if we remain existing name - we need to add 2 scenarios, the first with failed validation, the second - with passed

SinghAdes: We removed the logic of validation from invoiceLine , so i dont think faliure will occur in any case. So renaming it to suggested name.

SerhiiNosko: okay, you added failure scenario, thanks

SerhiiNosko (approved):

PR #858, opened 2023-02-08
MODEBSNET-53 added test with empty poline locations

grabsefx:
## Purpose
https://issues.folio.org/browse/MODEBSNET-53
## Approach
image


acquisitions/src/test/java/org/folio/EbsconetApiTest.java
```diff
@@ -38,6 +38,10 @@ void updateEbsconetOrderLineMixedFormat() {
   void cancelOrderLinesWithEbsconet() {
     runFeatureTest("cancel-order-lines-with-ebsconet");
   }
+  @Test
+  void updateEbsconetOrderLineEmptyLocations() {
+    runFeatureTest("update-ebsconet-order-line-empty-locations");
+  }
```
SerhiiNosko, line 44: Seems origninal karate test was not commited

acquisitions/src/main/resources/thunderjet/mod-ebsconet/features/update-ebsconet-order-line-empty-locations.feature
```diff
@@ -0,0 +1,143 @@
+Feature: Update Ebsconet Order Line
```
SerhiiNosko, line 1: Please add Jira ticket as a comment to easily map this test with the corresponding jira

SerhiiNosko (approved):

PR #868, opened 2023-02-14
[FAT-4438]-updated HoldingIds

SinghAdes:
## Purpose

## Approach

### TODOS and Open Questions

### Learning

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/features/item-and-holding-operations-for-order-flows.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/features/item-and-holding-operations-for-order-flows.feature
@@ -20,7 +20,7 @@ Feature: check Items and holding process.
     * def orderId = callonce uuid3
     * def poLineId = callonce uuid4
     * def initialInstanceId = globalInstanceId1
-    * def initialHoldingId = globalHoldingId1
+    * def initialHoldingId = globalHoldingId3
 
 
   Scenario: Create finances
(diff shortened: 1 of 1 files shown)


SinghAdes:
image

acquisitions/src/main/resources/thunderjet/mod-orders/features/item-and-holding-operations-for-order-flows.feature
```diff
@@ -20,7 +20,7 @@ Feature: check Items and holding process.
     * def orderId = callonce uuid3
     * def poLineId = callonce uuid4
     * def initialInstanceId = globalInstanceId1
-    * def initialHoldingId = globalHoldingId1
+    * def initialHoldingId = globalHoldingId3
```
SerhiiNosko, line 23: Please explain in reply to this commment and/or in PR description, why this change needed, maybe what corresponded story was performed that needed to update this karate tests etc.

SinghAdes: @SerhiiNosko :- In this test we are expecting Holding to be deleted and it was working as expected when we run this test individually  but we have another test where we are using same holdingId , 
so while performing GET for holdingId the expected code was 404 but since holding were created in other tests  so it was giving 200 instead of 404. 

I have changed the holding id and now it is working fine.

SerhiiNosko: I got it ,thanks. But it also makes this test not so stable, is it easy to not use global holdingId in this test?
Or that another test can it not delete global holding(and delete local holding itself). Maybe second option is preferable

SerhiiNosko: Now I got it, this test is new and written for recent story https://issues.folio.org/browse/MODORDERS-820 
And this test becomes failing randomly after implementing because it relates to results of another tests(holding1 and holding2), and it always passes when running individually.
One appraoch is to create new holding each time in this test or use global holding that not used in other tests

SerhiiNosko (approved):

Abdulkhakimov (approved):

PR #880, opened 2023-02-21
FAT-4575- Added permission

obozhko-folio:
[FAT-4575](https://issues.folio.org/browse/FAT-4575) - Karate test fail: [firebird/mod-bulk-operations] ModBulkOperationsApiTest {3}  firebird/mod-bulk-operations/features/holdings.feature

success


khandramai (changes requested):
Bulk-operations client shouldn't contain permissions from other modules. If tests fail without such permissions - this is a module problem. This should be fixed in the ModuleDescriptor not in karate tests.

PR #929, opened 2023-04-13
FAT-5355:Test request filtering by call number

MurodjonMukhamadaliev:
## Purpose
FAT-5355: https://issues.folio.org/browse/FAT-5355


mod-circulation/src/main/resources/vega/mod-circulation/features/requests.feature
```diff
@@ -1260,4 +1260,138 @@ Feature: Requests tests
...
+    Given path 'owners'
+    And request ownerEntityRequest
+    When method POST
+    Then status 201
+
+    # post a material type
+    * def extMaterialTypeId = call uuid1
+    * def extMaterialTypeId2 = call uuid1
+    * def extMaterialTypeName = 'Book'
+    * def extMaterialTypeName2 = 'Text'
```
alexanderkurash, line 1283: If there's variable named `something2`, there should probably also be `something1`, not just `something`. Also, I thought we only add prefix `ext` to those variables that come from somewhere outside the scenario, it means 'external'. In this case (and many others in this scenario) we're defining them here, so why are we using prefix `ext`?

mod-circulation/src/main/resources/vega/mod-circulation/features/requests.feature
```diff
@@ -1260,4 +1260,138 @@ Feature: Requests tests
...
     And retry until responseStatus == 422
     When method POST
-    And match $.errors[0].message == blockMessage
\ No newline at end of file
+    And match $.errors[0].message == blockMessage
+
+  Scenario: Test request filtering by call number
+    * def extHoldingsRecordId = call uuid1
+    * def extCallNumber = 'FAT-5355CN'
+    * def extCN = 'FAT5355CN'
```
alexanderkurash, line 1268: Why don't just name them callNumber1 and callNumber2?

mod-circulation/src/main/resources/vega/mod-circulation/features/requests.feature
```diff
@@ -1260,4 +1260,138 @@ Feature: Requests tests
...
+    And assert response.requests.length == 2
+    And match $.requests[0].item.callNumberComponents.callNumber == extCallNumber
+    And match $.requests[1].item.callNumberComponents.callNumber == extCN
+    And print response
+
+    Given path 'circulation/requests'
+    And param query = 'fullCallNumberIndex==' + itemPrefix + extCallNumber + itemSuffix
+    When method GET
+    Then status 200
+    And print response
```
alexanderkurash, line 1397: Still not checking the response, only the status

PR #1007, opened 2023-07-07
Draft to use existing tenant

bekhzodkurbonboev:
## Purpose
We need to run Karate tests against existing tenants.

## Approach
Introduce 2 new variables called:
- _'existingTenant'_ - name of the tenant which we need to run tests against;
- _'useExistingTenant'_ - flag to notify we are using existing tenant. If this variable is _true_ then all operations that changes tenant state (posting new tenant, enabling or disabling modules for tenant, ... ) will be skipped 

### TODOS and Open Questions
- [ ] Most of the modules use features in [common](https://github.com/folio-org/folio-integration-tests/tree/master/common/src/main/resources/common) to set up the environment/tenant, but not all. So need to look through all the test files to identify if there is a request to change tenant state directly (without using 'common');
- [ ] Most of the modules use global variables for set up, and there are constraints for uniqueness, e.g. unique_name, unique_id, ... . So after running one module any other subsequent test runs will fail in set up because of these duplicate values. 
image

 
### Learning

### DIFF
--- a/acquisitions/src/main/resources/karate-config.js
+++ b/acquisitions/src/main/resources/karate-config.js
@@ -6,8 +6,8 @@ function fn() {
   var env = karate.env;
 
   // The "testTenant" property could be specified during test runs
-  var testTenant = karate.properties['testTenant'];
-  var useExistingTenant = karate.properties['useExistingTenant'];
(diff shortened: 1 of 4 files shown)


bekhzodkurbonboev:
After setting necessary 'karate.properties' this worked: 
image
We need to set following 'karate.properties' in necessary environment (currently they are added here for testing): 
image

PR #1015, opened 2023-07-14
MODPUBSUB-270 scan collection to assert value for some tests

okolawole-ebsco:
## Purpose
Fix flaky situations with assertion of collections in a response


mod-circulation/src/main/resources/vega/mod-circulation/features/requests.feature
```diff
@@ -1271,7 +1272,7 @@ Feature: Requests tests
     # post a material type
     * def materialTypeId1 = call uuid1
     * def materialTypeId2 = call uuid1
-    * def materialTypeName1 = 'Book'
+    * def materialTypeName1 = 'Bookz'
```
okolawole-ebsco, line 1275: This change is weird.

this material type fails to be created because there is an existing material type called "book". mod-inventory-storage does a case insensitive comparison to existing material types.

The next material type "Text" ought to fail because there is an existing "text". It looks like something remove the existing material type of "text".

From what i can see these material types are not used anywhere else. I question why we would be testing if material types can be created in circulation tests. This should live in mod-inventory-storage. I am open to other recommendations on this.

roman-barannyk (approved):

PR #1025, opened 2023-07-26
FAT-6493: actualize match rules JSON according to SRM response

yaroslav-epam:
## Purpose
_To fix the Scenario: FAT-4701 check that default rules in karate tests matches what is in the SRM under test._

## Approach
_Use the actual JSON response from SRM. We see that test Scenario: FAT-4701... is not failing anymore:_
https://jenkins-aws.indexdata.com/job/Testing/job/Karate%20tests/674
image

### Learning
_[JIRA](https://issues.folio.org/browse/FAT-6493)_

### DIFF
--- a/data-import/src/main/resources/folijet/data-import/samples/samples_for_upload/default-marc-bib-rules.json
+++ b/data-import/src/main/resources/folijet/data-import/samples/samples_for_upload/default-marc-bib-rules.json
@@ -1,33 +1,33 @@
 {
   "001": [
     {
+      "rules": [
+      ],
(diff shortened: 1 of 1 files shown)


KaterynaSenchenko (commented):
Is only the formatting different, or some rules were not the same as well?

yaroslav-epam:
> Is only the formatting different, or some rules were not the same as well?

Looks like yes, after the `conditions` up to date in https://github.com/folio-org/folio-integration-tests/pull/1024

PR #1029, opened 2023-07-28
FAT-7117 fix search authority by lccn query params

ArtHov:
## Purpose
fix search authority by lccn query param nameing


mod-search/src/test/resources/spitfire/mod-search/search/authority-single-property-search.feature
```diff
@@ -117,7 +117,7 @@ Feature: Tests that authority searches by a single property
 
   Scenario Outline: Can search by lccn
     Given path '/search/authorities'
-    And param query = 'lccn="<value>"'
+    And param query = 'identifiers.value=="<value>"'
```
psmagin, line 120: lccn should work. if not we should find a reason in mod-search

PR #1060, opened 2023-08-30
[FAT-3315] - Enhance the Karate test cases for RTR implementation.

gurleenkaurbp:
## Purpose
The purpose of this PR is to adapt to the RTR implementation for the Karate Test cases.

## Approach
In the common login.feature file replace the /login end point with /login-with-expiry endpoint. The response will contain the cookies like folioAccessToken and folioRefreshToken. The folioAccessToken value can be then used as the x-okapi-token value in the request header for each scenario that runs. The login is called by every scenario before its execution as a background element in karate. The cookie value are stored in the variables cache and cleared, otherwise, the system will throw error as "Different tokens in Authentication and X-Okapi-Token. Use only one of them".

### TODOS and Open Questions

### Learning


pavankumar181 (approved):

Vignesh-kalyanasundaram (approved):

nirmalr-shah (approved):

steveellis (approved):
We could make it backwards compatible such that it checked the new endpoint, and then if 404 then checked the old one, but since the change is so small it will be easy to switch back if we have to. In any case it will be good to see how this works for the full test run so let's merge it and see what happens.

PR #1084, opened 2023-09-29
[MODCON - 85] - Fixed fail test cases in last scheduled karate test run 

azizbekxm:
## Purpose
https://issues.folio.org/browse/FAT-8087

## Approach
Fixed failed test case and increased permission length


SerhiiNosko:
All Karate tests passed:
image

SerhiiNosko (approved):

PR #1104, opened 2023-10-16
[FAT-7307] Update data-import tests for DITF

ncovercash:
# [Jira FAT-7307](https://issues.folio.org/browse/FAT-7307)

## Fixes

- [FAT-8955](https://issues.folio.org/browse/FAT-8955) (Setup failure after DITF merge)

## Fixes some of
- [FAT-6493](https://issues.folio.org/browse/FAT-6493) (Misc integration test failures)
  - [ ] 7835 `holdings issue`
  - [ ] 8588 `holdings issue`
  - [ ] 9088 `update via copycat issue`
- [FAT-7996](https://issues.folio.org/browse/FAT-7996) (EDI invoice failure since September)
  - [ ] 1131 `broken with Poppy changes`, [slack thread](https://folio-project.slack.com/archives/CBPA005DJ/p1697561677102659?thread_ts=1697518404.992089&cid=CBPA005DJ)

## Purpose
Update data-import Karate tests to support file splitting/S3 storage as implemented by the DITF.

## Approach
- Change module installation process
  - Karate relies on `mod-authtoken` being **disabled** to setup a tenant, create users, assign permissions, etc.  However, since `mod-data-import` now relies on `authtoken` (for the system user), we encountered a bit of a chicken/egg problem.
  - We followed the example of `mod-users-bl` to first setup only users/permissions modules, then perform the normal install steps (create users/etc).
  - Once the initial user creation was completed, we then can enable DI and related modules, and assign DI permissions.
- Add some missing permissions relating to `mod-entities-links` in Poppy
- Updated assertions of `mod-data-export` return values to support changes made in MDEXP-625 (https://github.com/folio-org/mod-data-export/blob/289c4af85cb128515275d36ec996736ca10e8991/ramls/schemas/jobExecution.json#L42-L62)
- Additional cleanup was done in many places, allowing reusable components to be better leveraged.
- Created/updated reusable components to support S3 uploads
  - Created a job execution query scenario based on S3 keys, to make it easier to find the `COMPOSITE_PARENT` job

## ✨Bonus✨

- Make created user's names reflect purpose (previously was hardcoded `Orders API Tests`)
- Fixed typo in tenant names (`testenant` -> `testtenant`)


KaterynaSenchenko (approved):

data-import/src/main/resources/folijet/data-import/global/common-data-import.feature
```diff
@@ -41,20 +40,37 @@ Feature: Do Data Import Using Parameters
     * def createDate = response.fileDefinitions[0].createDate
     * def uploadedDate = createDate
 
-    ## Upload marc-file
-    Given path 'data-import/uploadDefinitions', uploadDefinitionId, 'files', fileId
-    And headers headersUserOctetStream
-    * print 'FilePath ---', filePathFromSourceRoot
+    Given path 'data-import/uploadUrl'
```
RuslanLavrov, line 43: is split files functionality enabled on the karate env by default?

ncovercash: It will be going forward

data-import/src/main/resources/folijet/data-import/features/data-import-orders.feature
```diff
@@ -550,7 +552,6 @@ Feature: Test Data-Import holdings records
     And match response.orderFormat == "P/E Mix"
     And match response.purchaseOrderId == secondRecordOrderId
 
-  @Ignore
```
RuslanLavrov, line 553: as far as I remember this test has been ignored due to some issue related to "mix" open order creation. Does this test work now?

ncovercash: Yes, it was working in my testing

data-import/src/main/resources/folijet/data-import/global/common-data-import.feature
RuslanLavrov: will data import succeed in this case since the common-data-import.feature has been changed to load file using s3 but for file processing regular fileProcessor will be used? Also considering that many of the tests were refactored to re-use this feature file.

RuslanLavrov: yes, but since the functionality is not enabled the regular fileProcessor will be used that looks for file locally. Or did I miss something?

RuslanLavrov: so it is enabled at the deployment/cluster level for the environment/cluster where karate tests regular autorun is performed, right?

ncovercash: Correct

RuslanLavrov: thanks

PR #1142, opened 2023-12-06
MODFQMMGR-91: adding karate test for new entity type

kjain110:
https://issues.folio.org/browse/MODFQMMGR-91

Adding test for the new entity-type: Purchase Order Lines


mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/query.feature
```diff
@@ -245,6 +246,19 @@ Feature: Query
     * def totalRecords = parseInt(response.totalRecords)
     * assert totalRecords > 0
 
+#  Scenario: Run a query on the purchase order lines entity type
```
kjain110, line 249: Right now this test is commented out, need to validate it against the snapshot environment. I have merged my code 2 hrs back so by tom morning will be able to validate it

mweaver-ebsco (approved):

PR #1158, opened 2023-12-19
Rearrange mod-lists tests

ncovercash:
This separates the mod-lists Karate tests into separate features, for better granularity.  The scenarios themselves are left as-is, with the exception of:
- The `Post list, refresh list, export list, delete list` scenario (now in `delete.feature`) did not actually delete the list.  I added a request to delete the list.
- I extracted `update-list` into its own reusable feature


mweaver-ebsco (approved):
I like it :) This should make failures a bit easier to track down

kjain110 (approved):

PR #1169, opened 2024-01-02
Add validation Pick-up, borrowing-pickup, borrowing flow

gulzhaukhar-tukayeva-epam:
https://issues.folio.org/browse/FAT-9712

The Karate scenarios should cover below mentioned validations.
create transaction (POST) API have the following validations.
TransactionId should be unique for every transaction or else it will throw error 
Lending flow :
1) If the userId and barcode is not exist already, new user with type DCB will be created. 
2) Patron group should be validated at the time of user creation.
3) If it is a existing user and type is not dcb or shadow, error will be thrown.
4) Item needs to be present in inventory.(Real item)
Borrowing Flow/Borrowing-Pickup Flow :
1) If the userId and barcode is not exist already, error will be thrown. 
2) If the user exist but the type is DCB, error will be thrown
3) If the item barcode is already present in the inventory, error will be thrown.
4) If item is not present in inventory, new virtual item will be created.
5) If virtual item already exists, it will be reused. Make sure same id and barcode should be used.
6) Material type in the request should be present in inventory or else error will be thrown. If the material type is not given in the request, then we check for default material type as book in inventory, if it doesn't exist, we throw the error.
 
Pickup flow :
1) If the userId and barcode is not exist already, new user with type DCB will be created. 
2) Patron group should be validated at the time of user creation.
3) If it is a existing user and type is not dcb or shadow, error will be thrown.
4) If the item barcode is already present in the inventory, error will be thrown.
5) If item is not present in inventory, new virtual item will be created.
6) If virtual item already exists, it will be reused. Make sure same id and barcode should be used.
7) Material type in the request should be present in inventory or else error will be thrown. If the material type is not given in the request, then we check for default material type as book in inventory, if it doesn't exist, we throw the error.

Screenshot 2024-01-02 200851

### DIFF
--- a/common/src/main/resources/common/setup-users.feature
+++ b/common/src/main/resources/common/setup-users.feature
@@ -3,6 +3,7 @@ Feature: prepare data for api test
   Background:
     * url baseUrl
     * configure readTimeout = 3000000
+    * call pause 3000
     * configure headers = { 'Content-Type': 'application/json', 'Accept': 'application/json'  }
     * callonce login admin
     * def desiredPermissions = karate.get('desiredPermissions', [])
--- a/common/src/main/resources/common/tenant.feature
+++ b/common/src/main/resources/common/tenant.feature
@@ -2,7 +2,7 @@ Feature: Tenants
 
   Background:
     * url baseUrl
-    * configure retry = { count: 2, interval: 5000 }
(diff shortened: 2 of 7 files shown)


gulzhaukhar-tukayeva-epam:
Combine all flows

PR #1172, opened 2024-01-04
FAT-9722

TsaghikKhachatryan:
## Purpose
Fix the match failure

### Learning
https://issues.folio.org/browse/FAT-9722


TsaghikKhachatryan:
will be opened

PR #1195, opened 2024-01-19
[MODUSERS-421] - Disable username uniqueness validation for system users

azizbekxm:
## Purpose
https://issues.folio.org/browse/MODUSERS-421
## Approach
added two user one with system user type with same username, both should be stored. Other type users shouldn't

### DIFF
--- a/mod-consortia/src/main/resources/thunderjet/mod-consortia/features/consortia-user-type-update.feature
+++ b/mod-consortia/src/main/resources/thunderjet/mod-consortia/features/consortia-user-type-update.feature
@@ -156,5 +156,4 @@ Feature: Consortia User type Update tests
     """
     When method POST
     Then status 422
-# check error message
-  # User with this username already exists
+    And match response ==  {"errors":[{"message":"User with this username already exists","type":"1","code":"-1","parameters":[{"key":"username","value":"user2"}]}]}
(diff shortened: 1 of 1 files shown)


azizbekxm:
image

mod-consortia/src/main/resources/thunderjet/mod-consortia/features/consortia-user-type-update.feature
```diff
@@ -87,4 +87,73 @@ Feature: Consortia User type Update tests
...
+      "id":"b5fc7128-b783-4820-9a6b-3d4b35ebeff9",
+      "username": 'user1',
+      "active": true,
+      "personal": {"firstName":"User 1","lastName":'User 1'},
+      "type": "system"
+    }
+    """
+    When method POST
+    Then status 201
+
```
SerhiiNosko, line 127: please add here comment that we are creating 2 staff users where username validation accross tenants sshould exist

SerhiiNosko (approved):

SerhiiNosko (approved):

mod-consortia/src/main/resources/thunderjet/mod-consortia/features/consortia-user-type-update.feature
azizbekxm: Thanks, added

PR #1206, opened 2024-01-25
MODFQMMGR-125: Add deleted record test

bvsharp:
## Purpose
Add test for handling deleted records in mod-fqm-manager


mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/query.feature
```diff
@@ -274,6 +274,21 @@ Feature: Query
     * def totalRecords = parseInt(response.totalRecords)
     * assert totalRecords > 0
 
+  Scenario: Should return _deleted field to indicate that a record has been deleted
```
ncovercash, line 277: Nit: might be good to add the ticket number to the scenario name, so we can find where it came from later

ncovercash (approved):

PR #1228, opened 2024-02-05
Update field name in failing query test

bvsharp:
## Purpose
Update field name to fix failing query test


ncovercash (approved):
sure this couldn't be split into multiple PRs? a bit huge imo

PR #1233, opened 2024-02-07
profile picture CRUD operations karate test

gulzhaukhar-tukayeva-epam:
https://folio-org.atlassian.net/browse/FAT-11754

Karate test for POST, PUT, GET, DELETE profile picture 

image


SinghAdes (approved):

mod-users/src/main/resources/volaris/mod-users/features/profile-picture.feature
```diff
@@ -0,0 +1,91 @@
...
+    Given path '/users/configurations/entry/' + id
+    And request
+    """
+         {
+           "id": "#(id)",
+           "configName": "PROFILE_PICTURE_CONFIG",
+           "enabled": true,
+           "enabledObjectStorage": false,
+           "encryptionKey": "#(encryptionKey)"
+         }
```
gurleenkaurbp, line 32: Please write test cases when profile picture feature is not enabled.
When profile picture enabled for different kind of storage.
Please also write scenario to check the profile picture that is added is getting linked to users json, that scenario should be added to the existing users API. 
CC: @SinghAdes

Vignesh-kalyanasundaram (approved):

PR #1236, opened 2024-02-08
[MODINV-961] Modify test for instance soft delete to check if source record deleted

RomanChernetskyi:
## Purpose
Modify test for instance soft delete to check if source record deleted

## Approach
- Check record **state** to be DELETED
- Check to **suppressDiscovery** true
- Check **deleted** true
- Check **leaderRecordStatus** true


VRohach (approved):
@RomanChernetskyi  Awesome work!

Maksat-Galymzhan (approved):

RuslanLavrov (approved):

JavokhirAbdullayev (approved):

KaterynaSenchenko (approved):

PR #1238, opened 2024-02-09
Add Authtoken-Refresh-Cache = true to invildate permissions cache during login

SerhiiNosko:
Header name in mod-authtoken: https://github.com/folio-org/mod-authtoken/blob/c11c02504f3168f08266178e73ff25978da933be/src/main/java/org/folio/auth/authtokenmodule/MainVerticle.java#L23

How it uses: https://github.com/folio-org/mod-authtoken/blob/master/src/main/java/org/folio/auth/authtokenmodule/apis/FilterApi.java#L260


SerhiiNosko:
Locally tests are passing
image

azizbekxm (approved):

Dmitriy-Butramyou (approved):

PR #1242, opened 2024-02-14
[MODFISTO-432] Test updates for new transaction API and new parallel processing tests

damien-git:
## Purpose
[MODFISTO-432](https://folio-org.atlassian.net/browse/MODFISTO-432) - Implement logic to replace AllOrNothing to process list of transaction from request payload

## Approach
- Updated the tests to work with the new transaction API (without removing all old API calls)
- Added new parallel processing tests from an old PR

### TODOS
- Update all the tests to use the new transaction API

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/features/parallel-update-order-lines-same-order.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/features/parallel-update-order-lines-same-order.feature
@@ -1,0 +1,98 @@
+# For MODFISTO-260 and MODFISTO-432
+# This should be executed with at least 5 threads
+Feature: Update order lines for an open order in parallel
+
+  Background:
+    # This part is called once before scenarios are executed. It's important that all scenarios start at the same time,
+    # so all scripts must be called with callonce.
+    * url baseUrl
+    * callonce loginAdmin testAdmin
+    * def okapitokenAdmin = okapitoken
+    * callonce loginRegularUser testUser
+    * def okapitokenUser = okapitoken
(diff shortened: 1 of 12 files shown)


acquisitions/src/main/resources/thunderjet/mod-orders/features/parallel-update-order-lines-same-order.feature
```diff
@@ -0,0 +1,98 @@
...
+  Scenario: Update line 5
+    Given path 'orders/order-lines', poLineId5
+    When method GET
+    Then status 200
+    * def poLine = $
+    * set poLine.cost.listUnitPrice = 15
+    Given path 'orders/order-lines', poLineId5
+    And request poLine
+    When method PUT
+    Then status 204
```
azizbekxm, line 98: we can use Scenario Outline to avoid repition, there is two variable polinid and unitPrice

SerhiiNosko: it's only a suggestion, definetly we can remain as is

azizbekxm: if they doesn't work in scenario outline, you can resolve comment, it is okay to remain as it was

damien-git: Turns out it scenario outline works in parallel too. I updated it.

azizbekxm: just for knowladge sharing, I thought this part avoid parallel testing. do you think even if with this annotation scenario outline works parallel
image

SerhiiNosko (approved):

PR #1248, opened 2024-02-16
[MODGOBI-202] Create additional karate tests for mod-gobi module

Saba-Zedginidze-EPAM:
## Purpose
[https://folio-org.atlassian.net/browse/MODGOBI-202](https://folio-org.atlassian.net/browse/MODGOBI-202)

## Approach
Added tests to verify gobi integration:
* Make sure request is validated and changes are not made in case of invalid parameters
* Mappings should be added/modified only when they are present and further API calls should enforce this
* Orders should be affected by the changing mappings without any setbacks


acquisitions/src/main/resources/thunderjet/mod-gobi/features/gobi-api-tests.feature
```diff
@@ -80,3 +80,214 @@ Feature: mod-gobi api tests
     And headers { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': '*/*' }
     When method DELETE
     Then status 200
+
+  Scenario: Try to send invalid requests and check responses
+    * def valid_mapping = read('classpath:samples/mod-gobi/unlisted-print-serial-valid.json')
+    * def invalid_mapping_type = read('classpath:samples/mod-gobi/unlisted-print-serial-custom.json')
+    * def invalid_mapping_field = read('classpath:samples/mod-gobi/unlisted-print-serial-2.json')
+    * def invalid_mapping_translation = read('classpath:samples/mod-gobi/unlisted-print-serial-1.json')
```
SerhiiNosko, line 88: lets create new folder and move invalid mappings in that folder, it would be more clear to understand what mappings are not valid

acquisitions/src/main/resources/thunderjet/mod-gobi/features/gobi-api-tests.feature
```diff
@@ -80,3 +80,214 @@ Feature: mod-gobi api tests
...
+
+  Scenario: Make sure API methods behave correctly in different situations
+    * def valid_mapping = read('classpath:samples/mod-gobi/unlisted-print-serial-valid.json')
+
+    # Verify original mapping
+    Given path '/gobi/orders/custom-mappings/UnlistedPrintSerial'
+    And headers { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': '*/*' }
+    When method GET
+    Then status 200
+    And match response.orderMappings.orderType == 'UnlistedPrintSerial'
```
SerhiiNosko, line 137: let's add assertion that it is default mapping type
OrderMappingsView::mappingType

acquisitions/src/main/resources/thunderjet/mod-gobi/features/gobi-api-tests.feature
```diff
@@ -80,3 +80,214 @@ Feature: mod-gobi api tests
...
+    * def valid_mapping = read('classpath:samples/mod-gobi/unlisted-print-serial.json')
+
+    # Verify original mapping
+    Given path '/gobi/orders/custom-mappings/UnlistedPrintSerial'
+    And headers { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': '*/*' }
+    When method GET
+    Then status 200
+    And match response.orderMappings.orderType == 'UnlistedPrintSerial'
+    And match response.orderMappings.mappings[1].dataSource.default == "true"
+    And match response.orderMappings.mappings[2].dataSource.default == "false"
```
SerhiiNosko, line 139: not relates to this PR, but seems found issue in mod-gobi.
https://github.com/folio-org/mod-gobi/blob/master/src/main/java/org/folio/rest/service/GobiCustomMappingsService.java#L38
This method always loads default mappings and ignore custom mappings, we can fix it in scope of:
https://folio-org.atlassian.net/browse/MODGOBI-203

acquisitions/src/main/resources/thunderjet/mod-gobi/features/gobi-api-tests.feature
```diff
@@ -80,3 +80,214 @@ Feature: mod-gobi api tests
...
+    Then status 200
+
+    # Verify original mapping
+    Given path '/gobi/orders/custom-mappings/UnlistedPrintSerial'
+    And headers { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': '*/*' }
+    When method GET
+    Then status 200
+    And match response.orderMappings.orderType == 'UnlistedPrintSerial'
+    And match response.orderMappings.mappings[1].dataSource.default == "true"
+    And match response.orderMappings.mappings[2].dataSource.default == "false"
```
SerhiiNosko, line 197: in this and similar places we also can add assertions in these places if it custom or default to more easily distinguish mappings

acquisitions/src/main/resources/thunderjet/mod-gobi/features/gobi-api-tests.feature
```diff
@@ -80,3 +80,214 @@ Feature: mod-gobi api tests
...
+    And match $.poLines[0].orderFormat == 'Physical Resource'
+    And match $.poLines[0].poLineNumber == poLineNumber
+    And match $.poLines[0].titleOrPackage == 'Lightspeed Magazine'
+    And match $.poLines[0].vendorDetail.referenceNumbers[0].refNumber == '99974828470'
+
+    # Delete new mapping
+    Given path '/gobi/orders/custom-mappings/UnlistedPrintSerial'
+    And headers { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitoken)', 'Accept': '*/*' }
+    When method DELETE
+    Then status 200
```
SerhiiNosko, line 293: if it is possible it would be good to create new test for testing this functionality.
This LookupService is interacting with different services and that test would be very useful
https://github.com/folio-org/mod-gobi/blob/master/src/main/java/org/folio/gobi/LookupService.java

acquisitions/src/main/resources/thunderjet/mod-gobi/features/gobi-api-tests.feature
```diff
@@ -80,3 +80,214 @@ Feature: mod-gobi api tests
...
+    And request valid_mapping
+    When method POST
+    Then status 201
+
+    # Try to put an order for original mapping
+    Given path '/gobi/orders'
+    And headers { 'Content-Type': 'application/xml', 'x-okapi-token': '#(okapitoken)', 'Accept': '*/*' }
+    And request sample_po_original
+    When method POST
+    Then status 500
```
SerhiiNosko, line 233: if it is has some error code or message we can add assertion here.
It will also improve readability why test is failed in additional to testing error code provided

acquisitions/src/main/resources/thunderjet/mod-gobi/features/gobi-api-tests.feature
```diff
@@ -80,3 +80,272 @@ Feature: mod-gobi api tests
...
```
imerabishvili, line 236: **Question**
Why do we have 500 here? What errror exactly is returned from the server?

Saba-Zedginidze-EPAM: As the mapping gets updated, a new field becomes required to be present in the po, which is not included in sample_po_original

imerabishvili: But what is the server error then? Should we return 4xx instead?
`return 4xx as the problem is with the data being sent and not the server program. 5xx are reserved for server errors`
(c) Google

SerhiiNosko: we can improve this by adding proper status code in scope of next story https://folio-org.atlassian.net/browse/MODGOBI-203
The current story is only for covering module with more karate tests, changing mod-gobi logic is out of scope of this story

imerabishvili (approved):

azizbekxm (approved):

SerhiiNosko (approved):

PR #1267, opened 2024-02-28
[MODORDERS-1049] New test opening an order beyond restricted expenditures

damien-git:
## Purpose
[MODORDERS-1049](https://folio-org.atlassian.net/browse/MODORDERS-1049) - 500 error appears when trying to open order more than one time

## Approach
New test to open an order beyond restricted expenditures. Checks that inventory is not modified after failure and tries to open the order again after changing the amount.
Before [MODORDERS-1050](https://folio-org.atlassian.net/browse/MODORDERS-1050) is implemented this will check that the inventory rollback works when an unexpected error occurs while creating encumbrances.

## Related PR
[PR in mod-orders](https://github.com/folio-org/mod-orders/pull/847): Rollback inventory if open fails to create encumbrances

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
@@ -34,7 +34,7 @@ Feature: Open order failure with expenditure restrictions
     * def v = call createBudget { id: "#(budgetId)", fundId: "#(fundId)", allocated: 5 }
 
 
-  Scenario: Create an order and line
+  Scenario: Create an order and line going over budget
     * def v = call createOrder { id: "#(orderId)" }
     * def v = call createOrderLine { id: "#(poLineId)", orderId: "#(orderId)", fundId: "#(fundId)", listUnitPrice: 10, titleOrPackage: titleOrPackage }
 
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
@@ -1,0 +1,102 @@
+@parallel=false
+# Created for MODORDERS-1049
+# Related to MODORDERS-528 and open-order-failure-side-effects.feature
+Feature: Open order failure with expenditure restrictions
+
+  Background:
+    * print karate.info.scenarioName
+    * url baseUrl
+    * callonce loginAdmin testAdmin
+    * def okapitokenAdmin = okapitoken
+    * callonce loginRegularUser testUser
+    * def okapitokenUser = okapitoken
+    * def headersAdmin = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenAdmin)', 'Accept': 'application/json' }
+    * def headersUser = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenUser)', 'Accept': '*/*' }
+    * configure headers = headersUser
+
+    * callonce variables
+
+    * def createOrder = read('classpath:thunderjet/mod-orders/reusable/create-order.feature')
+    * def createOrderLine = read('classpath:thunderjet/mod-orders/reusable/create-order-line.feature')
+
+    * def ledgerId = callonce uuid1
+    * def fundId = callonce uuid2
+    * def budgetId = callonce uuid3
+    * def orderId = callonce uuid4
+    * def poLineId = callonce uuid5
+    * def titleOrPackage = callonce uuid6
+
+
+  Scenario: Create finances
+    * configure headers = headersAdmin
+    * def v = call createLedger { id: "#(ledgerId)", fiscalYearId: "#(globalFiscalYearId)", restrictEncumbrance: false, restrictExpenditures: true }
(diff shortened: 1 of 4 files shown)


acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
```diff
@@ -0,0 +1,102 @@
...
+  Scenario: Create finances
+    * configure headers = headersAdmin
+    * def v = call createLedger { id: "#(ledgerId)", fiscalYearId: "#(globalFiscalYearId)", restrictEncumbrance: false, restrictExpenditures: true }
+    * def v = call createFund { id: "#(fundId)", ledgerId: "#(ledgerId)" }
+    * def v = call createBudget { id: "#(budgetId)", fundId: "#(fundId)", allocated: 5 }
+
+
+  Scenario: Create an order and line
+    * def v = call createOrder { id: "#(orderId)" }
+    * def v = call createOrderLine { id: "#(poLineId)", orderId: "#(orderId)", fundId: "#(fundId)", listUnitPrice: 10, titleOrPackage: titleOrPackage }
```
SerhiiNosko, line 39: better to add comment that we are creating order line with bigger price that is allocated in budget to quicker understand why openning order failed

SerhiiNosko (approved):

acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
```diff
@@ -0,0 +1,102 @@
...
+    * def orderResponse = $
+    * set orderResponse.workflowStatus = 'Open'
+
+    Given path 'orders/composite-orders', orderId
+    And request orderResponse
+    When method PUT
+    Then status 422
+
+
+  Scenario: Check inventory records were not created during the failed open order operation
```
SerhiiNosko, line 56: inventory records were created initially, but they were rolled backed because of new logic, correct?
Better also to add comment about it, because it can be forgotten in some future

acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
```diff
@@ -0,0 +1,102 @@
...
+    * print 'Check instances'
+    Given path 'inventory/instances'
+    And param query = 'title==' + titleOrPackage
```
SerhiiNosko, line 59: currently we don't have possibility to delete instances, there is feature for next release to mark instances for deleting
https://folio-org.atlassian.net/browse/UXPROD-4303
Maybe order in the test was created without instance, since test is working

acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
damien-git: OK

acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
damien-git: That depends. Right now, before MODORDERS-1050 is implemented, we are checking that inventory records were rolled back (which is why I wanted to work on this ticket first). After MODORDERS-1050 is implemented this will simply check that validation prevented the creation of inventory records. Since MODORDERS-1050 is coming up very soon I did not want to mention the rollback in the test.

acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
damien-git: Right, I don't think any instance was created. This is just a sanity check that could be useful in the future.

acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
SerhiiNosko: okay got it now, and is it possible to test scenario separately when inventory rollback should be occured?

acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
SerhiiNosko: yes it makes sense

acquisitions/src/main/resources/thunderjet/mod-orders/features/open-order-failure-with-expenditure-restrictions.feature
damien-git: If we had a case this would be a bug (unless we cause some low-level problem in mod-finance-storage, but I am not sure we can rely on that in an integration test). It is safer to have initial validation in mod-oders make sure nothing bad happens before processing. I am not sure the rollback will work well in all cases. For instance suppose you open an order creating inventory automatically, then edit these records a bit, then unopen it without deleting the records, then reopen it and an error happens because of budgets. In this case the rollback would delete the inventory records which is probably not what the user would want.

Abdulkhakimov (approved):

PR #1268, opened 2024-02-29
order of response could differ causing flaky results

okolawole-ebsco:
## Purpose
fix flaky results in test

## Approach
Order of objects in HTTP response can differ, asserting the presence of values a different way


okolawole-ebsco:
@VRohach This particular test still fails locally when i execute it in the same run as other tests and not the data-import-integration test. The items created is 2 instead of 3. When I execute the test by itself, the number of items created is 3. This could point to some dependency to data-import-integration.

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #1269, opened 2024-03-01
EDGRTAC-81: Fix edge rtac scenarios

yusuf-murodov-epam:
## Purpose
After added new fields on rtac response, so updated edge-rtac responses. 

## Approach
- Added new match asserts in order to check new fields

image

### DIFF
--- a/edge-rtac/src/main/resources/karate-config.js
+++ b/edge-rtac/src/main/resources/karate-config.js
@@ -55,7 +55,7 @@ function fn() {
   if (env == 'snapshot-2') {
     config.baseUrl = 'https://folio-snapshot-2-okapi.dev.folio.org:443';
     config.edgeUrl = 'https://folio-snapshot-2.dev.folio.org:8000';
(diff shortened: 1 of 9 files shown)


edge-rtac/src/main/resources/karate-config.js
```diff
@@ -81,7 +82,7 @@ function fn() {
   } else if(env == 'folio-testing-karate') {
     config.baseUrl = '${baseUrl}';
     config.edgeUrl = '${edgeUrl}';
-    config.apikey = 'eyJzIjoiNXNlNGdnbXk1TiIsInQiOiJkaWt1IiwidSI6ImRpa3UifQ==';
+    config.apikey = 'eyJzIjoibTE2M0k2NTRHZ1pWOVBMdnRTa1MiLCJ0IjoiZGlrdSIsInUiOiJkaWt1X2FkbWluIn0K';
```
illia-borysenko-ebsco, line 85: Do we really need to update apiKey?

yusuf-murodov-epam: yeap, it is required

illia-borysenko-ebsco (approved):

illia-borysenko (approved):

PR #1281, opened 2024-03-11
[MODFQMMGR-192] Create automated test for entity type localization

ncovercash:


mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/entity-types.feature
```diff
@@ -131,4 +131,12 @@ Feature: Entity types
...
\ No newline at end of file
+    Then status 204
+
+  Scenario: Ensure all entity type and field names are localized
+    Given path 'entity-types'
+    When method GET
+    Then status 200
+
+    # calls this feature on each entity type summary
+    * call read('entity-type-localized.feature') response
```
ncovercash, line 142: This unintuitive line here is the best way to do a "for loop" in Karate; it will call `entity-type-localized.feature` on each element in the array `response`, taking the properties from each element (`id` and `label` in our case) and dumping them into the callee's scope.

kjain110 (approved):
LGTM

PR #1316, opened 2024-04-08
MODFQMMGR-223 Add gatling to the mod-fqm-manager tests

mweaver-ebsco:
This PR adds karate-gatling to mod-fqm-manager, based on the work done for the acquisitions module. It sets the shared submodule dependency versions in the parent POM's `dependencyManagemenet` section, so that consuming modules don't need to set a version in their `dependencies`; the version is set to `${project.version}`, so that we're always using the version that matches the root project's POM.

### DIFF
--- a/mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/query.feature
+++ b/mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/query.feature
@@ -13,6 +13,7 @@ Feature: Query
     * def orgContactEntityTypeId = '7a7860cd-e939-504f-b51f-ed3e1e6b12b9'
 
   Scenario: Post query
+    * print '## Create query'
     Given path 'query'
     And request { entityTypeId: '#(itemEntityTypeId)', fqlQuery: '{\"item_status\": {\"$in\": [\"missing\", \"lost\"]}}' }
     When method POST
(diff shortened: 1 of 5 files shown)


kjain110 (approved):

mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/query.feature
```diff
@@ -13,6 +13,7 @@ Feature: Query
     * def orgContactEntityTypeId = '7a7860cd-e939-504f-b51f-ed3e1e6b12b9'
```
mweaver-ebsco, line 13: @okolawole-ebsco, at the top of this file (~10 before the start of the diff of this file), we've got a `callonce login testUser`. The equivalent line in your acquisitions PR (#1300) changes the `callonce` to `call`. Should we do the same thing here?

okolawole-ebsco: It is only necessary if your scenario requires it. If multiple virtual users can execute the tests and everything looks green, no need to change it.

mod-fqm-manager/src/test/java/org/folio/QuerySimulation.scala
```diff
@@ -0,0 +1,41 @@
+package org.folio
+
+import com.intuit.karate.gatling.PreDef._
+import io.gatling.core.Predef._
+import org.apache.commons.lang3.RandomUtils
+
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+class OrderSimulation extends Simulation {
```
mweaver-ebsco, line 10: Clearly this was copy/pasted 😆 
```suggestion
class QuerySimulation extends Simulation {
```

ncovercash (approved):

PR #1317, opened 2024-04-08
FAT-12628 Implement simple DI test for Gatling execution

JavokhirAbdullayev:
## Purpose
[FAT-12628](https://folio-org.atlassian.net/browse/FAT-12628) Implement simple DI test for Gatling execution

## Approach
add scala test class
image

### DIFF
--- a/common/src/main/resources/common/destroy-data.feature
+++ b/common/src/main/resources/common/destroy-data.feature
@@ -4,7 +4,7 @@ Feature: destroy data for tenant
     * url baseUrl
     * configure readTimeout = 90000
     * configure retry = { count: 5, interval: 5000 }
-    * callonce login admin
+    * call login admin
 
   Scenario: purge all modules for tenant
 
--- a/common/src/main/resources/common/tenant.feature
+++ b/common/src/main/resources/common/tenant.feature
@@ -4,6 +4,7 @@ Feature: Tenants
     * url baseUrl
     * configure retry = { count: 2, interval: 5000 }
     * configure readTimeout = 3000000
+    * call login admin
 
   @create
   Scenario: createTenant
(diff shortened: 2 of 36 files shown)


common/src/main/resources/common/destroy-data.feature
```diff
@@ -4,7 +4,7 @@ Feature: destroy data for tenant
     * url baseUrl
     * configure readTimeout = 90000
     * configure retry = { count: 5, interval: 5000 }
-    * callonce login admin
```
okolawole-ebsco, line 7: why does this need to change? This is in the common module, so we have to be careful here.

common/src/main/resources/common/tenant.feature
```diff
@@ -4,6 +4,7 @@ Feature: Tenants
     * url baseUrl
     * configure retry = { count: 2, interval: 5000 }
     * configure readTimeout = 3000000
+    * call login admin
```
okolawole-ebsco, line 7: why this change?

data-import/src/test/java/org/folio/DataImportSimulation.scala
```diff
@@ -0,0 +1,41 @@
...
+    constantString + randomLong
+  }
+
+  val protocol = karateProtocol(
+    "/_/proxy/tenants/{tenant}" -> Nil,
+    "/_/proxy/tenants/{tenant}/modules" -> Nil,
+    "/_/proxy/tenants/{tenant}/install" -> Nil,
+    "/data-import/uploadDefinitions" -> Nil,
+    "/data-import/uploadDefinitions/{uploadDefinitionId}" -> Nil,
+    "/finance/transactions" -> Nil,
```
okolawole-ebsco, line 24: This endpoint called by file-upload.feature?

common/src/main/resources/common/destroy-data.feature
JavokhirAbdullayev: because it was throwing 401 error at clean-up process

common/src/main/resources/common/tenant.feature
JavokhirAbdullayev: image
 i've got these errors after running a test without these changes

JavokhirAbdullayev: I ran test against our perf rancher env

okolawole-ebsco (approved):

KaterynaSenchenko (approved):

Maksat-Galymzhan (approved):

PR #1325, opened 2024-04-13
FAT - 12652 Feature [folijet/data-import] DataImportApiTest {3} folijet/data-import/features/data-import-integration.feature

JavokhirAbdullayev:
## Purpose
Feature [folijet/data-import] DataImportApiTest {3} folijet/data-import/features/data-import-integration.feature
[FAT - 12652](https://folio-org.atlassian.net/browse/FAT-12652)

## Approach
fix issues

### DIFF
--- a/data-import/src/main/resources/folijet/data-import/features/data-import-integration.feature
+++ b/data-import/src/main/resources/folijet/data-import/features/data-import-integration.feature
@@ -28,9106 +28,9106 @@ Feature: Data Import integration tests
 
     * def samplePath = 'classpath:folijet/data-import/samples/'
     * def updateHoldings = 'data-import-integration.feature@UpdateHoldings'
-#
-#  Scenario: FAT-4701 check that default rules in karate tests matches what is in the SRM under test
-#    * def javaJsonUtils = Java.type('test.java.JsonUtils')
-#    Given path 'mapping-rules/marc-bib'
-#    And headers headersAdmin
-#    And def expectedJson = JSON.stringify(read('classpath:folijet/data-import/samples/samples_for_upload/default-marc-bib-rules.json'))
-#    When method GET
-#    Then status 200
-#    And def responseJson = JSON.stringify(response)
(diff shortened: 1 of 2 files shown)


Maksat-Galymzhan (approved):

data-import/src/main/resources/folijet/data-import/features/data-import-integration.feature
```diff
@@ -9239,7 +9239,7 @@ Feature: Data Import integration tests
     Then status 200
     And def overlayParsedRecord = response.parsedRecord
     And match overlayParsedRecord.content.fields[0] == parsedRecord.content.fields[0]
-    And match $overlayParsedRecord.content.fields[?(@.005)] != $parsedRecord.content.fields[?(@.005)]
+    And match $overlayParsedRecord.content.fields[?(@.005)] contains $parsedRecord.content.fields[?(@.005)]
```
RuslanLavrov, line 9242: why are we replacing it with the opposite verification? It is expected that 005 in the record after overlay is different from 005 in record before overlay.

JavokhirAbdullayev: oh good catch, thank you!

JavokhirAbdullayev: but test was failing on env

RuslanLavrov: it appears that this request returns the record before overlay (not after) because now GET metadata-provider/jobLogEntries endpoint returns record matchedId, but not recordId, and therefore `overlayParsedRecord` variable contains record before overlay, not after. That's why the test did not pass this verification. To fix this test we can use one of these endpoints to get the record after overlay process by received record matchedId:
GET /source-storage/source-records/{id}
GET /source-storage/records/{id}/formatted

JavokhirAbdullayev: thank you! fixed on new pr

PR #1340, opened 2024-04-26
[FAT-12699] - Create Gatling performance tests from karate

azizbekxm:
## Purpose
https://folio-org.atlassian.net/browse/FAT-12699

## Note
Other perforamnce testings are written that are in starting point in branch FAT-12699


acquisitions/src/main/resources/thunderjet/mod-organizations/features/acquisitions-api-tests.feature
```diff
@@ -6,18 +6,34 @@ Feature: Organizations API tests.
...
     * def headersUser = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenUser)', 'Accept': '*/*'  }
     * def headersAdmin = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenAdmin)', 'Accept': '*/*' }
 
     * configure headers = headersUser
 
-    * call variables
+    # Load global variables:
+    * callonce variables
+
+    # Define variables:
```
okolawole-ebsco, line 27: why are these commented out variables needed?

acquisitions/src/main/resources/thunderjet/mod-organizations/features/acquisitions-api-tests.feature
```diff
@@ -6,18 +6,34 @@ Feature: Organizations API tests.
     # uncomment below line for development
     #* callonce dev {tenant: 'testmodorgs'}
 
-    * call loginAdmin testAdmin
+    * callonce loginAdmin testAdmin
     * def okapitokenAdmin = okapitoken
+    * print okapitokenAdmin
```
okolawole-ebsco, line 11: was this line for debugging?

acquisitions/src/main/resources/thunderjet/mod-organizations/features/acquisitions-api-tests.feature
azizbekxm: removed, thanks

acquisitions/src/main/resources/thunderjet/mod-ebsconet/features/cancel-order-lines-with-ebsconet.feature
```diff
@@ -7,40 +7,40 @@ Feature: Cancel order lines with ebsconet
 
     * url baseUrl
 
-    * callonce login testAdmin
+    * call login testAdmin
```
SerhiiNosko, line 10: Here is reason why callonce changed to call:

lets say we have fundId one fundId for one db. If I want to create second fundId it will throw error.
Callonce call only time and generate uuid for example only one time. 
in Performance testing we use different users and repetable cycle with 10 times. So If I generate one uuid for all 10 tests it will throw in second cycle. 
 
For login it will genreate different token for different user everytime in every test. 
 
For different users we need different token and different uuids.

acquisitions/src/main/resources/thunderjet/mod-finance/features/allowable-encumbrance-and-expenditure-restrictions.feature
```diff
@@ -28,7 +28,7 @@ Feature: Test allowable encumbrance and expenditure restrictions
     * def poLineId = call uuid
     * def encumbranceId = call uuid
 
-    * call createFund { 'id': '#(fundId)', 'ledgerId': #(ledgerId) }
+    * def v = call createFund { 'id': '#(fundId)', 'ledgerId': '#(ledgerId)' }
```
SerhiiNosko, line 31: just for my understanding, why we need to assing result to some variable here?

azizbekxm: I wrote based on other PRs, it is working without it. That will be deleted

azizbekxm:
The karate files has logical changes are passing in local env too.
image
image
image

acquisitions/src/main/resources/thunderjet/mod-finance/features/allowable-encumbrance-and-expenditure-restrictions.feature
SerhiiNosko: we have dozens of similar places, thinking that we need to delete here a well to make code more clean

azizbekxm: The places introduced in my pr definitely helpful for the Performance testing. I tested with deleted this other similar places, tests failed because of the calling createFund.feature two times. I am not sure why outcome is different but this places helping performance testing working correctly.
image

SerhiiNosko (approved):

Saba-Zedginidze-EPAM (approved):

PR #1342, opened 2024-04-29
Remove unused vertx dependency

julianladisch:
No longer needed after the Vert.x code has been removed:
https://folio-org.atlassian.net/browse/FAT-153
https://github.com/folio-org/folio-integration-tests/pull/142


steveellis (approved):
This one has been open too long. Was hoping someone with more knowledge of this would pitch in but let's merge this and see how it goes.

PR #1359, opened 2024-05-20
[FAT-13426] Update Karate tests

BKadirkhodjaev:
## Purpose
- [FAT-13426](<https://folio-org.atlassian.net/browse/FAT-13426?atlOrigin=eyJpIjoiMTVhYjkwNzZkOTk4NGMzYzgyNmIwMWMyZDBkOGY3NTEiLCJwIjoiaiJ9>)

## Approach
- Add print to Background
- Refactor features using a single Scenario to remove parallel=false
- Use 4 default threads
- Use comments in place of removed scenarios
- Add missing modules
- Add indicators @Positive/@Negative
- Replace Examples with Tables by Removing Scenario Outline
- Add more reusable features
- Regenerate UUIDs

### Learning
- Refactored according to FAT-9581


damien-git:
Thanks. I don't know if you mean to do that as a later step, but you should replace typical operations like opening orders by the related reusable features. For instance use openOrder() instead of GET/PUT.

damien-git:
Also for the uuids (you might also plan to do it later, I don't know), we should replace "`* def fundId = callonce uuid1`" in Background by "`* def fundId = call uuid`" in the scenario. This makes it easier to move scenarios between features, or to add another scenario to the same feature without risking a conflict.

BKadirkhodjaev:
> Thanks. I don't know if you mean to do that as a later step, but you should replace typical operations like opening orders by the related reusable features. For instance use openOrder() instead of GET/PUT.

Yes I will also integrate your Jira comment into, probably already tomorrow.

acquisitions/src/main/resources/thunderjet/mod-orders/features/unopen-and-change-fund-distribution.feature
```diff
@@ -23,28 +23,29 @@ Feature: Unopen and change fund distribution
...
+  @Positive
+  Scenario: Unopen and change fund distribution
+    # 1. Create a fund and budget
     * configure headers = headersAdmin
     * call createFund { 'id': '#(fundId)', 'ledgerId': '#(globalLedgerId)'}
     * callonce createBudget { 'id': '#(budgetId)', 'fundId': '#(fundId)', 'allocated': 1000, 'statusExpenseClasses': [{'expenseClassId': '#(globalPrnExpenseClassId)','status': 'Active'}]}
 
 
-  Scenario: Create a composite order
+    # 2. Create a composite order
```
SerhiiNosko, line 34: lets use print instead of commenting .
This will help in troubleshouting when tests run on separate env and we could not run them locally.

SerhiiNosko: in this case we will be able to read full test flow from logs

damien-git:
Another thing: sometimes we use a URL for the ticket in a comment at the beginning. These URLs are deprecated and the links will probably stop working eventually (currently there is a redirect). We should update the URLs or just use the ticket name instead to avoid broken links in the future.

acquisitions/src/main/resources/thunderjet/mod-orders/features/unopen-and-change-fund-distribution.feature
azizbekxm: Comments are easier to read in cucumber report and code except for console logs.
image
using #
image

azizbekxm:
Can we add supporting of cucumber report for mod-orders. 
image

acquisitions/src/main/resources/thunderjet/mod-orders/features/unopen-and-change-fund-distribution.feature
damien-git: I don't know of a practical way to use cucumber reports when running tests locally (they don't appear in files like `overview-features.html`). It's also convenient to use the console log. Would adding an empty comment before the print help to separate the steps ? Or maybe a comment like "# Step" ?

azizbekxm: 1) It is not configured in acq-models, it is possible to implement. In mod-consortia last part there will be link to cucumber report. Example from mod-consortai tests https://github.com/folio-org/folio-integration-tests/pull/1359#issuecomment-2124120046
2) From my prespective title should be only in Scenario. In other places need description about flow what are we doing, what is situation with data. and why this step. With comments, it is easier to do so, with print more force to write just title

damien-git: @azizbekxm I would like us to implement this before we start changing prints into comments, because I would like to see how it changes my workflow. For instance I am currently able to click on a test and see the log right away while the other tests are still being executed - I am not sure I would be able to do that with cucumber reports.

azizbekxm: Of course, there is changes after implemention cucmber reports. You will no longer see the test title, there will be only one test title OrdersApiTest and link to cumumber report. To Run individual tests you should comment other one. In first attempt it is strange, but need to think. I am urging to move to cucumber report right now, we can think about that of course.

if not productive we can skip to implement cucumber because mod-orders has many tests that may make hard to test individual report. You can test this function running mod-consortia and there will be print.

Comments and Cucumber different suggestion and don't depend to each other. Even we are not moving to cucumber, we should still use comments instead of print. I head one advantage of print is to be able to search scenarios in logs, but error tells which line we have problem.

BKadirkhodjaev: So the consensus is to use print instead of a comment? I will implement the changes now

BKadirkhodjaev: Updated `cancel-order.feature`, and tried to include most recommendations

damien-git:
Regarding the new reusable features:
- These should not be reusable, they are too specific to particular tests: `check-budget-after-cancelling-order.feature`, `check-order-lines-after-cancelling-order.feature`. If possible we should keep that code in the features it came from (it is OK to have duplicated code in integration tests if it is not generic).
- `create-composite-order.feature` already exists, it is called `create-order.feature`.
- `create-order-lines.feature` already exists, it is called `create-order-line.feature`.

BKadirkhodjaev:
> @BKadirkhodjaev In the features you have modified, there are other places where reusable features can be used (for instance to open an order, or create an order line). I assume you intend to replace that too ?

Yep, currently `cancel-order.feature` is our proof-of-concept that will be propagated into other tests, starting with those already touched by me.

acquisitions/src/main/resources/thunderjet/mod-orders/features/cancel-order.feature
```diff
@@ -18,87 +16,61 @@ Feature: Cancel order
...
+    # Generate unique UUIDs
+    * def fundId = call uuid
+    * def budgetId = call uuid
+    * def orderId = call uuid
 
-
-  Scenario: Prepare finances
+    * print '1. Prepare finances'
     * configure headers = headersAdmin
     * call createFund { id: '#(fundId)' }
```
damien-git, line 34: Make sure to use "def v =" for all calls, even createFund and createBudget.

acquisitions/src/main/resources/thunderjet/mod-orders/reusable/create-order-lines.feature
```diff
@@ -0,0 +1,21 @@
+@ignore
+Feature: Create order lines
```
damien-git, line 2: There is already a feature to create an order line, it should be reused.

damien-git (changes requested):
Looks good, just 2 issues.

acquisitions/src/main/resources/thunderjet/mod-orders/reusable/create-order-lines.feature
BKadirkhodjaev: > There is already a feature to create an order line, it should be reused.

Previously I tried to adapt it, I will give it another try now

BKadirkhodjaev: Damien, I've adopted your solution but removed the table UUIsD to let the mod-orders do the id generation on its own:

image

SerhiiNosko (approved):

acquisitions/src/main/resources/thunderjet/mod-orders/features/cancel-order.feature
```diff
@@ -18,87 +16,61 @@ Feature: Cancel order
...
+    * def createOrder = read('classpath:thunderjet/mod-orders/reusable/create-order.feature')
+    * def createOrderLine = read('classpath:thunderjet/mod-orders/reusable/create-order-line.feature')
+    * def openOrder = read('classpath:thunderjet/mod-orders/reusable/open-order.feature')
+    * def cancelOrder = read('classpath:thunderjet/mod-orders/reusable/cancel-order.feature')
 
-    * def fundId = callonce uuid1
-    * def budgetId = callonce uuid2
-    * def orderId = callonce uuid3
+  @Positive
+  Scenario:
```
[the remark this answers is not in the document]
BKadirkhodjaev: Good catch, will fix now

acquisitions/src/main/resources/thunderjet/mod-orders/features/validate-fund-distribution-for-zero-price.feature
```diff
@@ -1,11 +1,9 @@
 # created for https://issues.folio.org/browse/MODORDERS-715
```
azizbekxm, line 1: Small suggestion to remove url. Use leave ticket number

azizbekxm (approved):

damien-git (approved):

PR #1407, opened 2024-06-20
Fixed testes after extracting credit from expended

azizbekxm:
## Purpose
MOFISTO-481
https://folio-org.atlassian.net/browse/FAT-14407
https://folio-org.atlassian.net/browse/FAT-14406
https://folio-org.atlassian.net/browse/FAT-14405
## Approach
Since, credit are being extracted from expenditures, calculation need to be fixed in karate tests. Credit transactions will no longer added to expenditures. Changed several places, before cancel invoice. After invoice verification still valid.

Extracting credit from expended amount will affect avaliable cashBalance, unavaliable amounts.


SerhiiNosko (approved):

damien-git:
Actually, separating credits from expenditures should not change available/unavailable/cashBalance. But maybe there are more tickets to implement before this is fixed.

PR #1419, opened 2024-06-24
[MODFISTO-488] New test cases when cancelling an invoice

damien-git:
## Purpose
[MODFISTO-488](https://folio-org.atlassian.net/browse/MODFISTO-488) - Wrong encumbrance when cancelling an invoice with an order that is not open

## Approach
- Added test cases when cancelling an invoice (in the cross-module tests)
- Updated cancelInvoice in invoice tests (without changing any logic)

[mod-finance-storage PR](https://github.com/folio-org/mod-finance-storage/pull/416)

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/cross-modules/features/cancel-invoice-with-encumbrance.feature
+++ b/acquisitions/src/main/resources/thunderjet/cross-modules/features/cancel-invoice-with-encumbrance.feature
@@ -1,4 +1,4 @@
-# MODINVOICE-544
+# For MODINVOICE-544 and MODFISTO-488
 Feature: Cancel an invoice with an Encumbrance
 
   Background:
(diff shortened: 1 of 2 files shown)


acquisitions/src/main/resources/thunderjet/cross-modules/features/cancel-invoice-with-encumbrance.feature
```diff
@@ -156,3 +160,176 @@ Feature: Cancel an invoice with an Encumbrance
...
+    * def fundId = call uuid
+    * def budgetId = call uuid
+    * def orderId = call uuid
+    * def poLineId = call uuid
+    * def invoiceId = call uuid
+    * def invoiceLineId = call uuid
+
+    * print "1. Create finances"
+    * configure headers = headersAdmin
+    * call createFund { id: "#(fundId)" }
```
BKadirkhodjaev, line 177: I think you forgot `def v = ...` for these calls

damien-git: Right, I thought we had updated this feature already and I just blindly copy-pasted that. Fixed.

BKadirkhodjaev (approved):

SerhiiNosko (approved):

PR #1422, opened 2024-06-25
[FAT-13846] Updated tests to separate credits from expenditures

damien-git:
## Purpose
[FAT-13846](https://folio-org.atlassian.net/browse/FAT-13846) - Update tests to separate credits from expenditures

## Approach
- Updated amounts and totals wherever credits are used (otherwise the tests are still valid)
- Added some test cases with credits to check totals
- I did not check if the tests pass because the implementation is not complete. So there might be something to fix afterwards, but for now it can be used as a target to achieve with the implementation.


SerhiiNosko (approved):

azizbekxm (approved):

azizbekxm:
re-enabled disabled karate tests related to credits feature in scope of this pr

PR #1486, opened 2024-08-08
[MODORDERS-1117]. Update new permission

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1117>

## Approach

- Update initialization features with new `orders-storage.po-lines-batch.collection.put` permission

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/orders-junit.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/orders-junit.feature
@@ -39,20 +39,21 @@ Feature: mod-orders integration tests
 
 
     * table userPermissions
-      | name                                   |
-      | 'orders.all'                           |
-      | 'finance.all'                          |
-      | 'inventory.all'                        |
-      | 'tags.all'                             |
-      | 'orders.item.approve'                  |
-      | 'orders.item.reopen'                   |
-      | 'orders.item.unopen'                   |
-      | 'invoice.all'                          |
-      | 'audit.all'                            |
-      | 'orders-storage.claiming.process'      |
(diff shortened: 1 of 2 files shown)


acquisitions/src/main/resources/thunderjet/mod-orders/orders-junit.feature
```diff
@@ -39,20 +39,21 @@ Feature: mod-orders integration tests
...
+      | 'orders.item.approve'                          |
+      | 'orders.item.reopen'                           |
+      | 'orders.item.unopen'                           |
+      | 'invoice.all'                                  |
+      | 'audit.all'                                    |
+      | 'orders-storage.claiming.process'              |
+      | 'inventory-storage.instances.item.get'         |
+      | 'inventory-storage.items.item.get'             |
+      | 'orders-storage.titles.item.get'               |
+      | 'orders-storage.po-lines-batch.collection.put' |
```
azizbekxm, line 56: I wonder maybe this permission is not need. these permission usually need when calling individual api  for example  trying to get individual item in scenarios to verify details let's say, so we need | 'inventory-storage.items.item.get'     |.

The case we don't verify item details, but api will use this fetch inventory items api inside of process then, we don't need the permission to fetch items because it is already exists in ModuleDescriptor for individual api under modulePermission array. 

As a conclusion, this permission maynot need unless we use the permission as scenario.

BKadirkhodjaev: The Karate tests do work without it, plus it is explicitly used by in the new associated PR: 
image

PR #1507, opened 2024-08-26
[MODORDERS-1167]. POL statuses remain "Ongoing" after unopen ongoing order

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/MODORDERS-1167>

## Approach

- Add 2 test cases for One-Time and Ongoing order types verifying that both Payment and Receipt statuses revert to Pending after unopening

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/features/open-and-unopen-order.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/features/open-and-unopen-order.feature
@@ -48,7 +48,6 @@ Feature: Open and unopen order
     Given path 'orders/order-lines', poLineId
     When method GET
     Then status 200
-
     And match response.paymentStatus == 'Awaiting Payment'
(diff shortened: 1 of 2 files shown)


acquisitions/src/main/resources/thunderjet/mod-orders/features/open-and-unopen-order.feature
```diff
@@ -0,0 +1,106 @@
...
+    * def v = call createOrderLine { id: '#(poLineId)', orderId: '#(orderId)', fundId: '#(fundId)' }
+
+    * print '4. Open the order'
+    * def v = call openOrder { orderId: '#(orderId)' }
+
+    * print '5. Check order line status after opening'
+    Given path 'orders/order-lines', poLineId
+    When method GET
+    Then status 200
+
```
Saba-Zedginidze-EPAM, line 51: Probably would be more readable to remove line breaks between `Then` and  `And match` statements

PR #1520, opened 2024-09-05
Fix failing mod-lists scenario

bvsharp:


mod-lists/src/main/resources/corsair/mod-lists/features/get-list-contents.feature
```diff
@@ -35,7 +35,6 @@ Feature: Scenarios that are primarily focused around getting list contents
     When method GET
     Then status 200
     And match $.content[0] == lastItem
-    And match $.content[1] == '#notpresent'
```
bvsharp, line 38: The line above checks that the offset functionality works correctly, so no need to check for the lack of a 2nd entry here.

ncovercash (approved):

PR #1521, opened 2024-09-05
[FAT-1626]. Add retries until the expected condition is met

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/FAT-16265>

## Approach

- Add retries to wait for 404 after the holdings are removed


acquisitions/src/main/resources/thunderjet/mod-orders/features/item-and-holding-operations-for-order-flows.feature
```diff
@@ -13,6 +13,8 @@ Feature: check Items and holding process.
     * def headersUser = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenUser)', 'Accept': '*/*'  }
     * configure headers = headersUser
 
+    * configure retry = { count: 10, interval: 10000 }
```
SerhiiNosko, line 16: lets do not more then 1 sec

SerhiiNosko (approved):

PR #1531, opened 2024-09-13
[FAT-16455]. Fix checkIn tests

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/FAT-16455>

## Approach

- Add missing purchase order and poLine creation
- Add correct poLineId in the request payload of the checkIn 
- Update status code assertion to user status code 400 instead 403

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/features/independent-acquisitions-unit-for-ordering-and-receiving.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/features/independent-acquisitions-unit-for-ordering-and-receiving.feature
@@ -529,7 +529,7 @@ Feature: Independent acquisitions unit for ordering and receiving
               locationId: "#(globalLocationsId)"
             }
           ],
-          poLineId: "#(pieceId)"
+          poLineId: "#(poLineId3)"
         }
       ],
       totalRecords: 1
@@ -561,7 +561,7 @@ Feature: Independent acquisitions unit for ordering and receiving
     When method DELETE
     Then status 204
 
-    # 7. Check piece 1 receivingStatus and update the piece should be forbidden
+    # 7. Check piece 1 receivingStatus and update the piece should return bad request
     * configure headers = headersUser
     Given path 'orders/pieces', pieceId
     When method GET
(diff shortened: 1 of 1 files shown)


acquisitions/src/main/resources/thunderjet/mod-orders/features/independent-acquisitions-unit-for-ordering-and-receiving.feature
```diff
@@ -561,7 +561,7 @@ Feature: Independent acquisitions unit for ordering and receiving
     When method DELETE
     Then status 204
 
-    # 7. Check piece 1 receivingStatus and update the piece should be forbidden
+    # 7. Check piece 1 receivingStatus and update the piece should return bad request
```
azizbekxm, line 564: I think update piece operation shouldn't be allowed. is bad request include some permission/authroziation exception, need to verify

BKadirkhodjaev: It has this exception, no sure if the piece payload is correct though: 
```16:29:39.747 [main] DEBUG com.intuit.karate - response time in milliseconds: 81
11 < 400
11 < Content-Type: application/json
11 < transfer-encoding: chunked
{
  "errors": [
    {
      "message": "There may be a link to either the holding or the location",
      "code": "mayBeLinkToEitherHoldingOrLocationError",
      "parameters": [
      ]
    }
  ],
  "total_records": 1
}
```

azizbekxm: Need to send correct payload and need to check 403 response. We can't change the error response code that is against main goal of test. I found that this error coming from[ this line](https://github.com/folio-org/mod-orders/blob/7a1c9694554f45c3685cc5778ab4d5c92b4a6445/src/main/java/org/folio/service/pieces/validators/PieceValidatorUtil.java#L29-L31) when payload has both location and holdingId, maybe we can remove either one to reach restriction check https://github.com/folio-org/mod-orders/blob/744be5bc5809576d25c50aefcc0687b6ade16079/src/main/java/org/folio/service/pieces/flows/create/PieceCreateFlowManager.java#L44

SerhiiNosko (approved):

azizbekxm (approved):

PR #1533, opened 2024-09-16
[FAT-16041] - Extend Central Ordering Karate tests coverage[Part 2]

azizbekxm:
## Purpose
https://folio-org.atlassian.net/browse/FAT-16041

## Approach
- Open order that contains 50 locations, half of them in tenant A, half of them in tenant B 
- Open order that contains 10 po lines, each have 5 locations with Affilation for tenant A or tenant B

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature
+++ b/acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature
@@ -13,22 +13,18 @@ Feature: Open order with many locations from different tenants
     * def fundId = callonce uuid1
     * def budgetId = callonce uuid2
 
-    * def orderId = call uuid
-    * def poLineId = call uuid
-    * def titleId = call uuid
-
-    * def poLineUuid = '5ee65782-ab71-4e07-9561-c400e3004a'
     * def centralLocationsId = '6ee65782-ab71-4e07-9561-c400e3004a'
     * def universityLocationsId = '7ee65782-ab71-4e07-9561-c400e3004a'
-    * def locationsCode = 'LOC'
 
 
+  @Positive
   Scenario: Prepare data: create fund and budget, and locations
+    * def locationsCode = 'LOC'
 
     * def v = call createFund { 'id': '#(fundId)' }
     * def v = call createBudget { 'id': '#(budgetId)', 'allocated': 100, 'fundId': '#(fundId)' }
 
-    ## Create 25 locations in central tenant
+    # Create 25 locations in central tenant
     * def centralLocations = []
     * def createCentralParameterArray =
       """
@@ -51,7 +47,7 @@ Feature: Open order with many locations from different tenants
 
     * def v = call createLocation centralLocations
 
-    ## Create 25 locations in university tenant
+    # Create 25 locations in university tenant
     * configure headers = { 'Content-Type': 'application/json', 'Authtoken-Refresh-Cache': 'true', 'x-okapi-token': '#(okapitoken)', 'x-okapi-tenant': '#(universityTenant)', 'Accept': 'application/json' }
     * def universityLocations = []
     * def createUniversityParameterArray =
@@ -74,8 +70,10 @@ Feature: Open order with many locations from different tenants
 
     * def v = call createLocation universityLocations
 
-
+  @Positive
   Scenario: Create Open 'ongoing' order and Verify Instance, Holdings and items in each tenant
+    * def orderId = call uuid
+    * def poLineId = call uuid
 
     ## 1.1 Create order
     * def v = call createOrder { id: '#(orderId)'}
@@ -182,8 +180,10 @@ Feature: Open order with many locations from different tenants
     Then status 200
     And match $.totalRecords == 25
 
-
+  @Positive
   Scenario: Open order that contains 10 po lines, each have 5 locations with Affilation for tenant A or tenant B
+    * def orderId = call uuid
+    * def poLineUuid = '5ee65782-ab71-4e07-9561-c400e3004a'
 
     ## 1. Create order
     * def v = call createOrder { id: '#(orderId)'}


azizbekxm:
image

acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature
```diff
@@ -0,0 +1,309 @@
...
+  Scenario: Prepare data: create fund and budget, and locations
+    * def v = call createFund { 'id': '#(fundId)' }
+    * def v = call createBudget { 'id': '#(budgetId)', 'allocated': 100, 'fundId': '#(fundId)' }
+
+    ## Create 25 locations in central tenant
+    * def centralLocations = []
+    * def createCentralParameterArray =
+      """
+      function() {
+        for (let i = 10; i < 35; i++) {
```
SerhiiNosko, line 34: we start from 10 due to some reasons?

SerhiiNosko (approved):

acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature
azizbekxm: removed two number from custom uuid to get different uuid. to match correct uuid length, started from 10 instead 1

SerhiiNosko: probably we can use random uuids here, or not?

acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature
```diff
@@ -0,0 +1,314 @@
...
+    * configure retry = { count: 5, interval: 1000 }
+
+    * callonce variables
+    * callonce variablesCentral
+    * callonce variablesUniversity
+
+    * def fundId = callonce uuid1
+    * def budgetId = callonce uuid2
+
+    * def orderId = call uuid
```
BKadirkhodjaev, line 16: Damien would've recommended to move all UUID generation in each Scenario for uniqueness & also can add @Positve/@Negative indicators for each Scenario

acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature
azizbekxm: Good idea, but it makes harder to keep track of uuids. For example, these locations is being used in poLine a later scenarios. https://github.com/folio-org/folio-integration-tests/blob/6f6b3dfa344eb75709df5d8dca98a5f56d724447/acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature#L216 .

SerhiiNosko: we do not use locationIds in assertions, so can have just 2 lists, 1 - with locations in cental tenant, 2 - with locations in university and uuids in this case do not matter. But if it hard to implement we can leave as it

BKadirkhodjaev: > Good idea, but it makes harder to keep track of uuids. For example, these locations is being used in poLine a later scenarios.
> 
> https://github.com/folio-org/folio-integration-tests/blob/6f6b3dfa344eb75709df5d8dca98a5f56d724447/acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature#L216
> 
> .

how about you pre-generate all on the Scenario level and just foreach the prepare list?

acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature
azizbekxm: it make sense, added notation and moved uuid generation while keeping uuids that will be used in each scenario. Thanks

acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature
azizbekxm: > we do not use locationIds in assertions, so can have just 2 lists, 1 - with locations in cental tenant, 2 - with locations in university and uuids in this case do not matter. But if it hard to implement we can leave as it

couldn't understand, we need locationIds for poLine a later.
1. prepare data scenario create locations in each tenant https://github.com/folio-org/folio-integration-tests/pull/1533/files#diff-7edc541c7b72f3478cf7f833e957c760aa9ad2e90097aaed8b0e7b352f42f650R50-R71
2.  verification scenarios. These locations will be set to poLines by using ids 
https://github.com/folio-org/folio-integration-tests/pull/1533/files#diff-7edc541c7b72f3478cf7f833e957c760aa9ad2e90097aaed8b0e7b352f42f650R86-R123
https://github.com/folio-org/folio-integration-tests/pull/1533/files#diff-7edc541c7b72f3478cf7f833e957c760aa9ad2e90097aaed8b0e7b352f42f650R86-R123
So we need uuid of created locations

sorry if I don't understand your solution correctly

azizbekxm: > > Good idea, but it makes harder to keep track of uuids. For example, these locations is being used in poLine a later scenarios.
> > https://github.com/folio-org/folio-integration-tests/blob/6f6b3dfa344eb75709df5d8dca98a5f56d724447/acquisitions/src/main/resources/thunderjet/consortia/features/prf-open-order-with-many-locations-from-different-tenants.feature#L216
> > 
> > .
> 
> how about you pre-generate all on the Scenario level and just foreach the prepare list?

I assume solution move about generate locations for each scenarios. if it is correct, the problem with this, it makes increase duplication and make it hard to follow the flow. So separated generation of locations

SerhiiNosko: > > we do not use locationIds in assertions, so can have just 2 lists, 1 - with locations in cental tenant, 2 - with locations in university and uuids in this case do not matter. But if it hard to implement we can leave as it
> 
> couldn't understand, we need locationIds for poLine a later.
> 
> 1. prepare data scenario create locations in each tenant https://github.com/folio-org/folio-integration-tests/pull/1533/files#diff-7edc541c7b72f3478cf7f833e957c760aa9ad2e90097aaed8b0e7b352f42f650R50-R71
> 2. verification scenarios. These locations will be set to poLines by using ids
>    https://github.com/folio-org/folio-integration-tests/pull/1533/files#diff-7edc541c7b72f3478cf7f833e957c760aa9ad2e90097aaed8b0e7b352f42f650R86-R123
>    https://github.com/folio-org/folio-integration-tests/pull/1533/files#diff-7edc541c7b72f3478cf7f833e957c760aa9ad2e90097aaed8b0e7b352f42f650R86-R123
>    So we need uuid of created locations
> 
> sorry if I don't understand your solution correctly

I thought we can iterate via locations array when adding location ids to po line, in this case we just need to get locationId from Location object. But if it is not possible so please ignore my comment

BKadirkhodjaev (approved):

PR #1585, opened 2024-10-24
Update edge-fqm test to match the updated API

mweaver-ebsco:


ncovercash (approved):

kjain110 (approved):

edge-fqm/src/main/resources/corsair/edge-fqm/features/edge-entity-types.feature
```diff
@@ -80,6 +80,6 @@ Feature: Entity types
     And path 'entity-types'
     When method GET
     Then status 200
-    * def baseResponse = response.entityTypes
+    * def baseResponse = response
```
mweaver-ebsco, line 83: I don't think this is enough. All this change does is make it so the next line will get tripped up by a new shape of data :P

ncovercash: ah yeah, may want to remove line 84

ncovercash: or change it. i remember something funky about the way the `missingPermissions` was handled (iirc mod-fqm-manager included `missingPermissions: null` and edge didn't?)

mweaver-ebsco: I'm thinking: go with the simple `And match response == edgeResponse` and see what happens. Worse case, it chokes on the `missingPermissions` and I can fix it easily enough

PR #1651, opened 2024-12-23
FAT-17756: set loadReferenceData: true for linked-data tests

pkjacob:
## Purpose
 set loadReferenceData: true for linked-data tests
 
 Some tests were failing because reference data like mod_of_issuance were not present in the DB. To fix this, I am loading all the ref data required by different modules by setting {loadReferenceData: true} in karate-config.js file. As a result, we can now remove the reference data that we were programmatically creating earlier


mod-linked-data/src/main/resources/karate-config.js
```diff
@@ -12,6 +12,7 @@ function fn() {
   var testTenant = karate.properties['testTenant'];
 
   var config = {
+    tenantParams: {loadReferenceData: true},
```
pkjacob, line 15: Setting `loadReferenceData: true` will initialize all the reference data for the tenant when the tenant is initialized.

So, we do not have to load the authority files and instance types programmatically, as we did earlier.

askhat-abishev (approved):

PR #1662, opened 2025-01-13
[FAT-17236] - Add Karate tests for FY finance bulk edit functionality

azizbekxm:
## Purpose
https://folio-org.atlassian.net/browse/FAT-17236

## Approach
-    Check finance-data endpoint that it returns data for all available 
-    Make some data restricted by acq unit and check that these data is skipped from finance-data 
 -   Check negative case like using multiple fiscalYearIds into the same 
 -   Check negative case that for new funds/budgets that do not exists in DB error is 
 -   Check negative case with providing incorrect allocation 
 -   Perform successful bulk edit operation, check that funds and budgets are updated and allocation transactions are 
 -   Check that dry run mode performs correctly, but without updated in DB for funds, budgets, 
 -   Check that action logs are appeared for both successful and failed cases
  -  Check that that each log contains correct state of bulk update in for updates in past


acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
```diff
@@ -0,0 +1,368 @@
...
+    * callonce login testAdmin
+    * def okapitokenAdmin = okapitoken
+    * callonce login testUser
+    * def okapitokenUser = okapitoken
+    * def headersUser = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenUser)', 'Accept': 'application/json' }
+    * def headersAdmin = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenAdmin)', 'Accept': 'application/json' }
+
+    * callonce variables
+
+    * def fundId1 = callonce uuid1
```
BKadirkhodjaev, line 17: Maybe better to have these IDs regenerated for each Scenario?

acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
```diff
@@ -0,0 +1,368 @@
...
+    When method GET
+    Then status 200
+    And match response.totalRecords == 0
+
+    # 2. Verify get finance data details
+    Given path 'finance/finance-data'
+    And param query = 'fiscalYearId==' + fiscalYearId2
+    When method GET
+    Then status 200
+    And match $.fyFinanceData[0].fiscalYearId == fiscalYearId2
```
BKadirkhodjaev, line 167: Would be better with `And match each $.fyFinanceData[*] ..` but it is more of a nitpick

acquisitions/src/main/resources/thunderjet/mod-finance/reusable/createFund.feature
```diff
@@ -9,6 +9,7 @@ Feature: fund
     * def ledgerId = karate.get('ledgerId', globalLedgerId)
     * def externalAccountNo = karate.get('externalAccountNo', '1111111111111111111111111')
     * def fundStatus = karate.get('fundStatus', 'Active')
+    * def acqUnitIds = karate.get('acqUnitIds', [])
```
BKadirkhodjaev, line 12: Did you check to see whether this addition does not break other tests?

acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
azizbekxm: this scenario is more about checking exact field exists or not rather than inside of data, so current style is more appropriate. I used `And match each $.fyFinanceData[*] ` this approach in other places

acquisitions/src/main/resources/thunderjet/mod-finance/reusable/createFund.feature
azizbekxm: yes if other places didn't use this param, it will get default [] empty value
image

acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
azizbekxm: there is many finance data creation in background section, if we move to each scenario it increases line of code. so keeping in one place and later use in each scenario approach is more good options

BKadirkhodjaev: Okay that makes sense, thanks

BKadirkhodjaev (approved):

Saba-Zedginidze-EPAM (approved):

PR #1695, opened 2025-02-10
FAT-18337 -  Common utility functions refactoring

viktor-draban:
In this PR, the functionality of the common module was adopted to work against Eureka.


khandramai:
@viktor-draban , please, don't remove existing code. You can create separate folder for eureka and put new code into it or create new features with postfix -eureka.

khandramai:
OKAPI-related classes and features which you reimplemented should be marked as deprecated.

testrail-integration/src/main/java/org/folio/test/TestBase.java
khandramai, line 1: Please, create TestBaseEureka.java instead of changing existing class.

khandramai: TestBase.java can be marked as @Deprecated and @Disabled.

khandramai (approved):

MaximKa99 (approved):
Everything is ok

PR #1698, opened 2025-02-10
[MODFIN-407] Finance data test updates

damien-git:
## Purpose
[MODFIN-407](https://folio-org.atlassian.net/browse/MODFIN-407) - Make fund active when creating or increasing allocation

## Approach
- Made the tests more parallel (previously update log checks would fail in parallel, and other checks with funds and budgets could also fail randomly)
- Added checks for `fundStatus` automatic update

### TODOS and Open Questions
- TODO: More validation in mod-finance, especially for `budgetCurrentAllocation`, see [MODFIN-403](https://folio-org.atlassian.net/browse/MODFIN-403). Right now the integration test will fail because a 422 is expected and a 500 is returned.

[mod-finance-storage PR](https://github.com/folio-org/mod-finance-storage/pull/449), [mod-finance PR](https://github.com/folio-org/mod-finance/pull/273)

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
@@ -251,7 +251,8 @@ Feature: Karate tests for FY finance bulk get/update functionality
     Given path 'finance/finance-data'
     And request requestBody
     When method PUT
-    Then status 422
+    # TODO: change to 422 after fixing validation in mod-finance
+    Then status 500
 
     Given path 'finance-storage/fund-update-logs'
     When method GET
--- a/acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
@@ -1,5 +1,5 @@
 Feature: Karate tests for FY finance bulk get/update functionality
-  # for FAT-17236
+  # for FAT-17236, MODFIN-402, MODFIN-407
 
   Background:
     * print karate.info.scenarioName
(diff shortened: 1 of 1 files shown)


SerhiiNosko (approved):
Lets merge in this case this PR after https://folio-org.atlassian.net/browse/MODFIN-403 to prevent test failures in the pipeline

SerhiiNosko:
> Lets merge in this case this PR after https://folio-org.atlassian.net/browse/MODFIN-403 to prevent test failures in the pipeline

@damien-git or you can fix mod-finance code to return 422 in scope of existed PR in mod-finance that allows to merge this karate PR

damien-git:
@SerhiiNosko if we're going to do a temporary change, it would be easier to check for a 500 in the test. Checking for properties in mod-finance will probably require getting the budget and other records from storage, and it makes sense to wait for MODFIN-403 for that.

acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
```diff
@@ -246,18 +246,19 @@ Feature: Karate tests for FY finance bulk get/update functionality
...
-    Then status 422
+    # TODO: change to 422 after fixing validation in mod-finance
+    Then status 500
 
     Given path 'finance-storage/fund-update-logs'
     When method GET
     Then status 200
-    And match $.totalRecords == 1
-    And match $.fundUpdateLogs[0].status contains 'ERROR'
+    * def log = karate.jsonPath(response, "$.fundUpdateLogs[?(@.jobDetails.fyFinanceData[0].fundId=='"+fundId1+"')]")[0]
```
BKadirkhodjaev, line 260: Bit hard to read & looks like something that may break some day, did the previous version not work as expected?

acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
```diff
@@ -308,14 +309,14 @@ Feature: Karate tests for FY finance bulk get/update functionality
     Given path 'finance-storage/fund-update-logs'
     When method GET
     Then status 200
-    And match $.totalRecords == 2
-    And match $.fundUpdateLogs[*].status contains 'COMPLETED'
-    And match $.fundUpdateLogs[*].jobDetails contains {"fyFinanceData":[{"fundId":"#(fundId1)","budgetId":"#(budgetId1)","fundCode":"FND1UPDATED","fundName":"Fund 1 Updated","fundTags":{"tagList":["updatedTag1"]},"budgetName":"Budget 1 Updated","fundStatus":"Inactive","budgetStatus":"Active","fiscalYearId":"#(fiscalYearId1)","fiscalYearCode":"FDATAFY2044","fundAcqUnitIds":[],"transactionTag":{"tagList":["Urgent"]},"fundDescription":"UPDATED subdivided by geographic regions, to match individual selectors","budgetAcqUnitIds":[],"budgetAfterAllocation":1100.0,"budgetAllocationChange":100.0,"transactionDescription":"End of year adjustment","budgetInitialAllocation":1000.0,"budgetCurrentAllocation":1000.0,"budgetAllowableEncumbrance":160.0,"budgetAllowableExpenditure":150.0}]}
+    * def log = karate.jsonPath(response, "$.fundUpdateLogs[?(@.jobDetails.fyFinanceData[0].fundId=='"+fundId1+"')]")[0]
```
BKadirkhodjaev, line 312: The same pattern here and on 2 subsequent lines. If it's too hard to optimize I am fine with it though.

BKadirkhodjaev (approved):

acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
damien-git: No, the different tests were not designed to work in parallel but it didn't have the `@parallel=false` tag. Part of the problem here is that the job number is not returned when you send the finance data update, and the ledger and fiscal year are not part of that table, so it is hard to find the right job to check the results automatically. With a unique `fundId` and looking at `jobDetails` we are able to find it. The `jsonPath` pattern is pretty common in the tests, but it's true it is not easy to read.

azizbekxm (approved):

PR #1704, opened 2025-02-13
[MODFISTO-517] Create budgets with batch allocations

damien-git:
## Purpose
[MODFISTO-517](https://folio-org.atlassian.net/browse/MODFISTO-517) - Create budget when adding batch allocation to fund without budget

## Approach
- Added new tests
- Fixed unrelated issue with error messages (related to first MODFIN-403 PR)

[mod-finance-storage PR](https://github.com/folio-org/mod-finance-storage/pull/450), [mod-finance PR](https://github.com/folio-org/mod-finance/pull/275)

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-finance/features/finance-data.feature
@@ -1,5 +1,5 @@
 Feature: Karate tests for FY finance bulk get/update functionality
-  # for FAT-17236, MODFIN-402, MODFIN-407
+  # for FAT-17236, MODFIN-402, MODFIN-407, MODFISTO-517
 
   Background:
     * print karate.info.scenarioName
@@ -23,10 +23,8 @@ Feature: Karate tests for FY finance bulk get/update functionality
     * def ledgerId3 = callonce uuid { n: 7 }
     * def fundId1 = call uuid
     * def fundId2 = call uuid
-    * def fundId3 = call uuid
     * def budgetId1 = call uuid
     * def budgetId2 = call uuid
-    * def budgetId3 = call uuid
 
     * def userId = "00000000-1111-5555-9999-999999999992"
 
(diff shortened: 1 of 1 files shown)


SerhiiNosko (approved):

azizbekxm (approved):

damien-git:
@azizbekxm I re-added "TODO: change to 422 after fixing validation in mod-finance" because this is not resolved. User errors should never result in a 500. In this case validation is incomplete, it should check the latest budget current allocation and return a 422 if it does not match what is being sent. In mod-finance the error "Allocation change cannot be greater than current allocation" is still using the budget current allocation from the user finance data to check the allocation change - it should use the real budget current allocation, not what the user is sending.

azizbekxm:
> @azizbekxm I re-added "TODO: change to 422 after fixing validation in mod-finance" because this is not resolved. User errors should never result in a 500. In this case validation is incomplete, it should check the latest budget current allocation and return a 422 if it does not match what is being sent. In mod-finance the error "Allocation change cannot be greater than current allocation" is still using the budget current allocation from the user finance data to check the allocation change - it should use the real budget current allocation, not what the user is sending.

Seems mistunderstanding here. This part is intended to get 500 error and check ERROR logs are being actually saved. The error we will get this
4 < 500
4 < Date: Wed, 12 Feb 2025 12:55:49 GMT
4 < Content-Type: application/json
4 < Transfer-Encoding: chunked
4 < Connection: keep-alive
{
  "errors": [
    {
      "message": "Failed to create allocations",
      "code": "failedToCreateAllocations",
      "parameters": [
        {
          "key": "cause",
          "value": "Expenditure restriction does not allow this operation"
        }
      ]
    }
  ],
  "total_records": 1
}

Also we thought about to add validation but I saw big two validation method  https://github.com/folio-org/mod-finance-storage/blob/55f597fab4dce85d3943c9ca264a4d785e6e522a/src/main/java/org/folio/service/transactions/batch/BatchTransactionChecks.java#L215 not increase complexity by copying these two, for now skipped this and get error from batch transaction and record as error logs.

PR #1710, opened 2025-02-18
MODSOURCE-859: Create new API POST /source-storage/records/{id}/un-delete

AndreiBordak:


mod-inventory/src/main/resources/folijet/mod-inventory/features/inventoryFeatureTest.feature
```diff
@@ -297,6 +297,19 @@ Feature: inventory
     And match response.leaderRecordStatus == "d"
     And match response.parsedRecord.content.leader.charAt(5) == "d"
 
+    Given path 'source-storage/records' + recordId + '/un-delete'
```
pkjacob, line 300: Let us change the description of the scenario (line 269) to `Soft delete and un-delete instance marc type`

pkjacob, line 300: I'm not sure if this is the right place for this test.

The purpose of all tests in this file is to validate mod-inventory's interaction with other modules. For example, in line 280, the `/mark-deleted` API of mod-inventory is called, and the result is verified in subsequent lines. This approach ensures that mod-inventory's interaction with mod-source-record-storage is tested.

However, this test directly invokes a POST API on mod-source-record-storage, bypasing mod-inventory. That do not match with other tests in this file. Instead, we should wait for MODINV-1139 to be implemented and then validate the new behavior by invoking the PUT instance API.

Optionally, you can also consider moving this test to the "mod-source-record-storage" directory.

AndreiBordak: I think we can add entire E2E test later, I've added IT for new endpoint in source-records-storage module, so I think it's enough for now.

askhat-abishev:
@AndreiBordak looks like changes are made to other team's module. Maybe it makes sense to add that team's members as a reviewers to this PR?

mod-inventory/src/main/resources/folijet/mod-inventory/features/inventoryFeatureTest.feature
AndreiBordak: Added integration test for /un-delete endpoint in mod-source-record-storage. For E2E flow we'll have Karate test:
https://github.com/folio-org/folio-integration-tests/pull/1722

PR #1712, opened 2025-02-18
Implement Automated Testing Code from POC Branch for ET

kjain110:


mod-fqm-manager/src/main/resources/karate-config.js
```diff
@@ -82,8 +82,8 @@ function fn() {
     config.baseUrl = 'https://folio-perf-corsair-okapi.ci.folio.org:443';
     config.admin = {
       tenant: 'fs09000000',
-      name: 'admin',
-      password: 'bugfest09'
+      name: 'diku_admin',
+      password: 'admin'
     };
     config.prototypeTenant = 'fs09000000';
```
bvsharp, line 88: Can we update the rancher config to something that will work again, since perf is gone? Maybe dev-corsair with diku_admin as the prototype tenant?

mod-fqm-manager/src/test/java/org/folio/ModFqmManagerTest.java
```diff
@@ -35,6 +35,7 @@ static List<Arguments> features() {
     return Stream
       .of(
         "entity-types",
+              "validate-entity-type",
```
bvsharp, line 38: nit: formatting

bvsharp (approved):

mweaver-ebsco (approved):

PR #1735, opened 2025-03-05
Odin module is adopted

viktor-draban:
- mod-batch-print tests are adopted to eureka env
- some common-eureka module tags have been renamed using camelcase
- updated createEntitlement functionality using queryParams by default and add ability to skip them
- added an additional check, if entitlementCreation returns a status cancelled\cancellation_failed\failed then stop execution


khandramai (approved):

common/src/main/resources/common/eureka/setup-users.feature
```diff
@@ -5,28 +5,37 @@ Feature: prepare data for api test
...
+    * if (typeof entitlementDefaultBehavior !== 'undefined' && entitlementDefaultBehavior == false) queryParam = {}
+    * call read('classpath:common/eureka/application.feature@applicationSearch')
     * def entitlementTamplate = read('classpath:common/eureka/samples/entitlement-entity.json')
     Given url baseUrl
     Given path 'entitlements'
+    And params queryParam
     And request entitlementTamplate
     When method POST
 
+    Given path 'entitlement-flows'
```
MaximKa99, line 26: i do not understand why we use here endpoint to check async operation if originally we call sync (because we do not specify async parameter) endpoint to create entitlements.

viktor-draban: Our first environment returned an immediate response about entitlement creation, so I added that code here. In that environment the request seems synchronous, but I can't rely on the subsequent environment and its behavior. Do you think we should leave it here or is it completely redundant and should be removed?

viktor-draban: It makes sense

viktor-draban: done

MaximKa99 (approved):

PR #1754, opened 2025-03-12
[MODFISTO-515] Delete a pending payment to revert creation

damien-git:
## Purpose
[MODFISTO-515](https://folio-org.atlassian.net/browse/MODFISTO-515) - Support deleting pending payments with batch transaction API

## Approach
- New tests for the batch API, to delete a pending payment
- Fixed and updated the tests to check the allowable encumbrance/expenditures (previously the test for expenditures was actually checking encumbrance; also it is more realistic to create a pending payment before creating a payment - this way the test will be less dependent on the implementation)

[mod-finance-storage PR](https://github.com/folio-org/mod-finance-storage/pull/465)

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-finance/features/allowable-encumbrance-and-expenditure-restrictions.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-finance/features/allowable-encumbrance-and-expenditure-restrictions.feature
@@ -1,17 +1,17 @@
 Feature: Test allowable encumbrance and expenditure restrictions
 
   Background:
+    * print karate.info.scenarioName
+
     * url baseUrl
-    # uncomment below line for development
-#    * callonce dev {tenant: 'testfinance4'}
     * call login testAdmin
     * def okapitokenAdmin = okapitoken
 
     * call login testUser
     * def okapitokenUser = okapitoken
 
-    * def headersUser = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenUser)', 'Accept': 'application/json'  }
-    * def headersAdmin = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenAdmin)', 'Accept': 'application/json'  }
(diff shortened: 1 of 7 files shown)


SerhiiNosko (approved):

acquisitions/src/main/resources/thunderjet/mod-finance/features/allowable-encumbrance-and-expenditure-restrictions.feature
```diff
@@ -84,6 +84,95 @@ Feature: Test allowable encumbrance and expenditure restrictions
       | negative   | 100        | 10           | 50         | 25          | 17           | 110          | 30     | 422    |
 
 
+  Scenario Outline: Test allowable encumbrance with pending payment: remaining encumbrance would be <remaining>
```
BKadirkhodjaev, line 87: Can this be change to use just the **Scenario** with the newer table-driven approach?

damien-git: I started implementing that, and then I realized this would not work well, because for each example you create a new fund, budget, encumbrance, and sometimes pending payment. So for instance you would have to create a table for funds with fundId1, fundId2 etc (because you need the fund id to create the related budget and encumbrance), and it would not be much more readable.

But also, I think there is nothing wrong with using `Scenario Outline` as it is done in this feature. We should avoid using it when the tests cannot be used in parallel and the different examples are meant to be executed one after the other, and we expect all of the data to be there afterwards. But if you create a scenario with different cases that are independant and the next scenario is not using that data, it makes sense to use Scenario Outline.

We can still use createXXX reusable features in this case to simplify the code.

damien-git: @BKadirkhodjaev I rewrote allowable-encumbrance-and-expenditure-restrictions.feature to use reusable features and add checks for codes. I will merge it, but you might find the changes interesting.

PR #1772, opened 2025-03-18
FAT-19191: Fix order of classification number in API response

pkjacob:
Fix order of classification number in API response

In GET /resource/{instanceId} API response, dewey classification comes before lc classification
In GET /resource/{workId} API response, lc classification comes before dewey classification


askhat-abishev:
@pkjacob can we check the response using different approach?

AndreiBordak (approved):

PR #1785, opened 2025-03-21
[FAT-19390]. Fix incorrect series used in FY

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/FAT-19390>

## Approach

- Change FY series back to TESTFYA with codes TESTFYA0001 and TESTFYA0002


damien-git (changes requested):
This fixes test execution with a single thread, but I think you should also revert the removal of "`@parallel=false`", because the test is written assuming scenarios are executed in order. The `@parallel=false` tag can help us find features to update, it also makes sense to use it given Karate's default parallel execution.

damien-git (approved):

BKadirkhodjaev:
Merging with 1 BE approval

PR #1866, opened 2025-04-24
[MODORDERS-1265]. Remove ISBN validation/normalization logic

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODORDERS-1265


azizbekxm (approved):

damien-git (changes requested):
You probably have to edit `poline_change_instance_connection.feature`.

SerhiiNosko:
> You probably have to edit `poline_change_instance_connection.feature`.

Thanks, both tests were verified locally
attempt_to_update_unmod_order
instance_connection

SerhiiNosko:
Also now we do not set quialifier at all in mod-orders code, so removed it from tests as well.
Now behaviour is consistent with Inventory application that does not have qualifier also, it has only productId and type
image

damien-git (approved):

PR #1876, opened 2025-04-28
[MODORDERS-1272] Added ability to run the tests locally with eureka-cli

damien-git:
## Purpose
[MODORDERS-1272](https://folio-org.atlassian.net/browse/MODORDERS-1272)

## Approach
- Fixed application deployment when there is no dependency array (needed with eureka-cli)
- Added a config parameter for the keycloak client id, since it is different by default for eureka-cli
- Fixed configs
- Merged recent commits from master branch

### TODOS and Open Questions
- **TODO before merge**: define `kcClientId`/`kcClientSecret` for other teams' `karate-config.js`
- TODO: make it work on rancher and snapshot
- TODO: add this env variable to mod-login-keycloak in the default eureka-cli config: `KC_ADMIN_TOKEN_TTL: 59s`
- Make the tests use a testUser user as before, avoid using an admin user for everything (in other PRs)
- TODO: use a consistent login method (currently `common/login` or `eureka/login`).


SerhiiNosko (approved):

BKadirkhodjaev:
I need to try it locally, please do not merge it as it looks like some of the stuff here is only intended for local runs

common/src/main/resources/common/eureka/application.feature
```diff
@@ -27,7 +27,7 @@ Feature: Applications
     * def uniqueAppIds = karate.distinct(appIds)
 
     # search dependencies among found applications, extract dependencies names, remove dependencies name duplicates
-    * karate.forEach(uniqueAppIds, appId => dependencyNames.push(appDescriptions.filter(descriptor => descriptor.id == appId).flatMap(x => x.dependencies.map(dep => dep.name))))
+    * karate.forEach(uniqueAppIds, appId => dependencyNames.push(appDescriptions.filter(descriptor => descriptor.id == appId).flatMap(x => x.dependencies ? x.dependencies.map(dep => dep.name) : [])))
```
BKadirkhodjaev, line 30: This is needed because in local Eureka environment, in the default profile we don't have any dependencies as we have in other envs, which are typically divided into app-minimal and app-complete that needs app-minimal to be installed first

- <https://github.com/folio-org/eureka-setup/blob/master/eureka-cli/config.combined.yaml#L9>

So without this addition Karate locally with Combined profile will fail

BKadirkhodjaev (approved):

damien-git:
@SerhiiNosko I meant to finish this before merging (from the TODO list):
- TODO before merge: define kcClientId/kcClientSecret for other teams' karate-config.js
I noticed that was missing yesterday, I should have turned that PR back into a draft...

I will try to finish that quickly today.

PR #1879, opened 2025-05-05
[MODORDERS-1272] Fixed permissions for mod-finance tests

damien-git:
## Purpose

## Approach
Fixed permissions for mod-finance tests.  The test user has only mod-finance permissions. The admin user has only non-finance permissions - this way it won't be used unnecessarily.
The current common features only support the creation of a single user (with okapi the test and admin users were created at the same time). After using `setup-users.feature` to create the test user, `finance-junit.feature` calls some `setup-users.feature` scenarios again to create the admin user.

### TODOS and Open Questions
- TODO (later): move some tests to cross-modules or modify them so they don't use invoice or orders permissions.
- TODO (later): fix `finance.feature` to support eureka (if possible)
- TODO (soon): fix permissions for other acquisitions tests, using a similar separation between admin and user permissions - for instance in mod-orders tests the test user will have only mod-orders permissions.


acquisitions/src/main/resources/thunderjet/mod-finance/features/budgets-totals-calculation.feature
```diff
@@ -1,15 +1,14 @@
...
 
   Background:
+    * print karate.info.scenarioName
     * url baseUrl
-    # uncomment below line for development
-    #* callonce dev {tenant: 'testfinance1'}
-    * callonce login testAdmin
-    * def okapitokenAdmin = okapitoken
-    * def headersAdmin = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenAdmin)', 'Accept': 'application/json', 'x-okapi-tenant': '#(testTenant)'  }
-    * def headersUser = headersAdmin
```
BKadirkhodjaev, line 10: Are you sure we don't need admin here for budget/fund stuff?

BKadirkhodjaev (approved):

PR #1880, opened 2025-05-06
[MODORDERS-1272] Fixed permissions for mod-orders tests

damien-git:
## Purpose

## Approach
- Updated tests to use a test user and an admin user with different permissions. The test user has all permissions for mod-orders, the admin user has all the other permissions.
- Modified `independent-acquisitions-unit-for-ordering-and-receiving.feature`  to create and use a new user, `testUser2`, which has some specific permissions (previously it was using the admin user to edit orders, but the new users do not allow that)
- Introduced 2 new functions, `createAdditionalUser` and `getUserIdByUsername`. Updated mod-finance tests to use `getUserIdByUsername` too.

### TODOS and Open Questions
- TODO: move some tests to mod-invoice or cross-modules.
- TODO: convert the remaining tests

### DIFF
--- a/acquisitions/src/main/resources/global/orders.feature
+++ b/acquisitions/src/main/resources/global/orders.feature
@@ -2,7 +2,6 @@ Feature: global orders
 
   Background:
     * url baseUrl
-    * call login testAdmin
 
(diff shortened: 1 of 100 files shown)


SerhiiNosko (approved):

BKadirkhodjaev (approved):
These are some great changes

PR #1885, opened 2025-05-15
[MODORDERS-1272] Fixed users and permissions for consortia tests

damien-git:
## Purpose

## Approach
- Cleaned up consortia-orders-junit.feature to make it easier to read
- Fixed the headers for all requests in consortia tests (note that "`And header`" does not work when `configure` is used)
- Removed unused permissions

### TODOS
- Continue fixing tests for Eureka
- Look at non-junit features
- Someone else should look at the remaining errors in `open-order-with-locations-from-different-tenants` and `reopen-and-change-instance-connection-order-with-locations-from-different-tenants`, they are probably caused by errors in the consortia tests (as multiple errors were cancelling each other and the user/permissions were fixed)


SerhiiNosko (approved):
nice that we have option to run consortia karate tests now on Eureka at least on rancher

BKadirkhodjaev (approved):

PR #1886, opened 2025-05-15
Enhance capability retrieval by tracking found and missing permissions

okolawole-ebsco:
## Purpose
Show missing capabilities to help troubleshooting

## Approach
Tracking existing capabilities and return capabilities that do not exists in the "expected" list


damien-git:
This could be very useful.

PR #1900, opened 2025-05-29
Move other features under one master feature for parallel exec

okolawole-ebsco:
- Most of the feature are moved verbatim but there are tweaks to get to a passing state. The previous version rely on many magic variables which causes troubleshooting hell.
- There is a karate quirk where feature files executed with `call read(...)` ignore all tags in the feature file, including `@ignore`. This means that reusable functions should declared outside the main feature files so that they don't get executed.

## Purpose
Move other features under one master feature for parallel exec


okolawole-ebsco:
Going in a different direction to support parallel execution and individual execution better.

PR #1913, opened 2025-06-12
Load capability sets using paging approach

pfilippov-epam:
## Purpose
Load capability sets using a paging approach to avoid issue: 414 Request-URI Too Long

## Approach
- Refactor load capability sets action to avoid `Request-URI Too Long`

### TODOS and Open Questions

### Learning


BKadirkhodjaev (approved):
Thanks, this looks like a great addition

PR #1931, opened 2025-06-27
[FAT-20154] Separate @ignore scenarios from the features they are called from

Saba-Zedginidze-EPAM:
## Purpose
[[FAT-20154] Separate @ignore scenarios from the features they are called from](https://folio-org.atlassian.net/browse/FAT-20154)

## Approach
- Extract reusable scenarios from test feature files into separate reusable features
- Extract test-specific reusable scenarios into helper features
- Combine all commonly used reusables and use them to get rid of code duplication

## Screenshots
mod-invoice
image

mod-organizations
image

mod-orders (only affected features)
image

cross-modules (only affected features)
image


BKadirkhodjaev (approved):
Good one

SerhiiNosko (approved):

PR #1966, opened 2025-07-29
[MODORDERS-1299|1300] Add a new feature test for instance link change with different settings

Saba-Zedginidze-EPAM:
## Purpose
[[MODORDERS-1300] Holdings duplicated in new instance after selecting "Create new" during changing instance connection for P/E mix order](https://folio-org.atlassian.net/browse/MODORDERS-1300)
[[MODORDERS-1299] After changing instance in P/E mix order, new instance is not displayed and locations show "Invalid reference"](https://folio-org.atlassian.net/browse/MODORDERS-1299)

## Approach
Add new scenario for changing instance and moving/creating holdings for POL

### Screenshots
image
image


acquisitions/src/main/resources/thunderjet/mod-orders/reusable/create-piece-with-holding.feature
```diff
@@ -1,6 +1,6 @@
 @ignore
 Feature: Create piece with holding id
-  # Parameters: pieceId, poLineId, titleId, holdingId, receivingTenantId, format
+  # Parameters: id*, poLineId*, titleId*, holdingId, receivingTenantId, format, createItem
```
BKadirkhodjaev, line 3: What do the asterisk here signify?

Saba-Zedginidze-EPAM: Just to mark the required arguments for the reusable

Saba-Zedginidze-EPAM: I thought we marked them with *'s, good catch, will modify

PR #1978, opened 2025-08-06
[MODFQMMGR-888] Fix FQM consortia tests

bvsharp:
## Purpose
[MODFQMMGR-888](https://folio-org.atlassian.net/browse/MODFQMMGR-888): Investigate and fix FQM karate failures


kjain110 (approved):

mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/consortia/consortia-fqm.feature
```diff
@@ -1,3 +1,4 @@
+@smoke
```
bvsharp, line 1: will remove

PR #2032, opened 2025-09-26
FAT-21885: Include sourceVersion in PUT request payload to avoid optimistic lock exception

pkjacob:
## Purpose
Include sourceVersion in PUT request payload to avoid opti…

### DIFF
--- a/mod-linked-data/src/main/resources/citation/mod-linked-data/features/authority/samples/authority_person_update.json
+++ b/mod-linked-data/src/main/resources/citation/mod-linked-data/features/authority/samples/authority_person_update.json
@@ -76,5 +76,6 @@
       "userId": "#(userId)"
     }
   },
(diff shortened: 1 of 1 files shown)


mod-linked-data/src/main/resources/citation/mod-linked-data/features/authority/samples/authority_person_update.json
```diff
@@ -76,5 +76,6 @@
       "userId": "#(userId)"
     }
   },
-  "relatedRecordVersion": 0
+  "relatedRecordVersion": 0,
+  "sourceVersion":0
```
pkjacob, line 80: It is ok to hardcode sourceVersion as "0". This is a record that we created in the previous test step. Nobody else is going to update it.

zepheiryan (approved):

PR #2116, opened 2025-12-05
[FAT-22638]. Add a shared instance to persist case ids in memory once

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/FAT-22638>

## Approach

- Add classes to initialize a singleton shared instance that will store case ids in memory once after startup
- This instance will live in the lifetime of the Maven test execution and will significantly speed up Test Rail result creation
- This will move case id retrieval from the per test-class mode to once at Maven test phase
- Test locally on environment deployed with Eureka CLI and with a temporary main method that spawns multiple threads

> Test Rail being disabled (with no env vars)

image

> Test Rail being enabled

image


testrail-integration/src/main/java/org/folio/test/services/TestRailService.java
```diff
@@ -27,51 +26,24 @@ public class TestRailService {
...
-      }
-      logger.info("cacheCaseIds:: Offset: {}, next: {}", responsePayload.offset(), responsePayload.links().next());
-      requestOffset++;
-    }
-
-    logger.info("cacheCaseIds:: Cached {} case ids", cachedCaseIds.size());
   }
 
   public void createResults(Integer runId, Map<String, Results> results) {
+    var caseIds = SharedInstanceInitializer.getSharedInstance().getCaseIds();
```
SerhiiNosko: name a bit confuses with shared instances functionality that we implemented recently.
Can rename to SharedCacheInitializer.getSharedCache()

SerhiiNosko (approved):

PR #2275, opened 2026-04-14
[FAT-22677] Implement Karate test for the test case C353543 Update po lines when an order is closed with the "Cancelled" reason

Saba-Zedginidze-EPAM:
## Purpose
[[FAT-22677] Implement Karate test for the test case C353543 Update po lines when an order is closed with the "Cancelled" reason](https://folio-org.atlassian.net/browse/FAT-22677)

image

### DIFF
--- a/acquisitions/src/test/java/org/folio/OrdersExtendedApiTest.java
+++ b/acquisitions/src/test/java/org/folio/OrdersExtendedApiTest.java
@@ -80,7 +80,7 @@ class OrdersExtendedApiTest extends TestBaseEureka {
   }
 
   @Test
-  @DisplayName("(Thunderjet) (@C353543) Update po lines when an order is closed with the 'Cancelled' reason")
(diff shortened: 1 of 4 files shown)


SerhiiNosko (approved):

acquisitions/src/test/java/org/folio/OrdersExtendedApiTest.java
```diff
@@ -77,4 +78,11 @@ void pieceStatusTransitionsClaiming() {
   void addPieceToCancelledOngoingOrder() {
     runFeatureTest(Feature.FEATURE_2.getFileName());
   }
+
+  @Test
+  @DisplayName("(Thunderjet) (@C353543) Update po lines when an order is closed with the 'Cancelled' reason")
```
BKadirkhodjaev: A small change: `@C353543` -> `C353543`

BKadirkhodjaev (approved):

PR #2281, opened 2026-04-17
[FAT-25489]. Implement Karate test for the test cases C987716 C987717 C987720

BKadirkhodjaev:
## Purpose

- <https://folio-org.atlassian.net/browse/FAT-25489>

## Approach

- Implement C987716, C987717 and C987720
- Add more fine-controls of tenant initialization and cleanup

### Ignore tenant destruction

> VM options to prevent automatic tenant destruction, for debugging purposes

image

### Destroy a particular tenant

> VM options to create a small ignorable test method, to delete a tenant; useful when manual tenant destruction is necessary

image

### DIFF
--- a/acquisitions/src/test/java/org/folio/EdgeOrdersApiTest.java
+++ b/acquisitions/src/test/java/org/folio/EdgeOrdersApiTest.java
@@ -75,7 +75,8 @@ class EdgeOrdersApiTest extends TestBaseEureka implements AcquisitionsTest {
     runFeature("classpath:common/eureka/destroy-data.feature");
   }
 
-  @Test@Override
+  @Test
+  @Override
   @DisplayName("(Thunderjet) Run features")
   @DisabledIfSystemProperty(named = "test.mode", matches = "no-shared-pool")
   public void runFeatures() {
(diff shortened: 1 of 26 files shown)


acquisitions/src/test/java/org/folio/shared/AcquisitionsTest.java
```diff
@@ -0,0 +1,11 @@
+package org.folio.shared;
+
+public interface AcquisitionsTest {
+  void beforeAll();
+
+  void afterAll();
+
+  default void destroyTenant() {}
```
BKadirkhodjaev, line 8: By default it is not required to be implemented, and is blank, but if necessary can be used to wire both the `@AfterAll` and the custom delete tenant test method

acquisitions/src/test/java/org/folio/EdgeOrdersApiTest.java
```diff
@@ -60,22 +61,24 @@ public EdgeOrdersApiTest() {
...
 
   @AfterAll
-  void edgeOrdersApiTestAfterAll() {
+  @Override
+  public void afterAll() {
     runFeature("classpath:common/eureka/destroy-data.feature");
   }
 
-  @Test
+  @Test@Override
```
SerhiiNosko: 2 annotations are on the same line

SerhiiNosko (approved):
very detailed nice tests

PR #2295, opened 2026-04-24
[MODORDERS-1439] Changed order status automatic update logic

damien-git:
## Purpose
[MODORDERS-1439](https://folio-org.atlassian.net/browse/MODORDERS-1439) - Cancelling an invoice does not update order correctly

## Approach
- Added a new test matching the scenario in the JIRA description
- Changed the tests related to automatic order status updates, using the new logic

## Related PRs
- [mod-orders](https://github.com/folio-org/mod-orders/pull/1291)
- [mod-invoice](https://github.com/folio-org/mod-invoice/pull/617)

### TODOS and Open Questions

### Learning


SerhiiNosko (approved):

acquisitions/src/main/resources/thunderjet/cross-modules/features/cancel-invoice-linked-to-order.feature
```diff
@@ -858,3 +858,101 @@ Feature: Cancel an invoice linked to an order
...
+    * def v = call payInvoice { invoiceId: '#(invoiceId)' }
+
+    * call pause 100
+
+    # 10. Check the order was closed automatically
+    * print "10. Check the order was closed automatically"
+    Given path 'orders/composite-orders', orderId
+    When method GET
+    Then status 200
+    And match $.workflowStatus == 'Closed'
```
BKadirkhodjaev: Lets wrap orders status check in a retry until

acquisitions/src/main/resources/thunderjet/mod-orders/helpers/helper-order-status-automatic-change.feature
```diff
@@ -0,0 +1,58 @@
...
+    * def v = call updateOrderLine { id: '#(poLineId)', paymentStatus: '#(initialPaymentStatus)', receiptStatus: '#(initialReceiptStatus)' }
+
+    # 4. Close order if initialWorkflowStatus is Closed
+    * if (initialWorkflowStatus == 'Closed') karate.call('classpath:thunderjet/mod-orders/reusable/close-order.feature', { orderId: orderId })
+
+    # 5. Check order's initial workflow status
+    Given path 'orders/composite-orders', orderId
+    When method GET
+    Then status 200
+    And match response.workflowStatus == initialWorkflowStatus
```
BKadirkhodjaev: Better to have retries for both order and poLine status. Here's an [example](https://github.com/folio-org/folio-integration-tests/commit/e0543052f09a885463aa11c306738b340011c57c#diff-b9ecc1c1f7a4fa797c0168b6230106aab29da8260e9fa611552dcfa329e76b43R37) for orders

acquisitions/src/main/resources/thunderjet/mod-orders/helpers/helper-update-po-lines-when-order-cancelled.feature
```diff
@@ -33,6 +33,8 @@ Feature: Helper for "update-po-lines-when-order-cancelled"
     When method PUT
     Then status 204
 
+    * call pause 100
```
BKadirkhodjaev, line 36: Might not be needed afterall

damien-git: @BKadirkhodjaev There was polling here in the previous version, instead of a sleep, and that was a bug. The order status was sometimes matching the expected status before the change, and it continued without an error. If the test had waited a bit, it would have noticed the status was not matching the expected one.

Polling is assuming you know the previous value (which might not be true), and it is also assuming the value will not switch twice (which could very well happen given the complexity of the mod-orders and mod-invoice workflows). Given that this caused the bug described above, I think we need to be very careful when replacing a sleep.

acquisitions/src/main/resources/thunderjet/mod-orders/helpers/helper-order-status-automatic-change.feature
damien-git: I am going to add a pause here (see below for explanation).

acquisitions/src/main/resources/thunderjet/cross-modules/features/cancel-invoice-linked-to-order.feature
damien-git: OK in this case.

acquisitions/src/main/resources/thunderjet/mod-orders/helpers/helper-order-status-automatic-change.feature
damien-git: nevermind, I guess a retry will work better here because we're not testing this part of the workflow, just the one after
Edit: something else is wrong here, a retry without pause would let it pass just before it changes value

damien-git: OK, I ended up removing 2 cases that could not be tested anyway, because of the auto-closing and reopening changing line statuses. So I was able to use retries here in the end.

BKadirkhodjaev (approved):

Saba-Zedginidze-EPAM (approved):

PR #2311, opened 2026-05-01
EUREKA-882: Add Karate coverage for mod-login-keycloak 

pkjacob:
**Tested APIs**

1. POST /authn/login
2. POST /authn/login-with-expiry
3. POST /authn/refresh
4. POST /authn/logout
5. POST /authn/logout-all
6. POST /authn/update
7. POST /authn/credentials
8. DELETE /authn/credentials
9. GET /authn/credentials-existence
10. POST /authn/password-reset-action
11. GET /authn/password-reset-action/{actionId}
12. POST /authn/reset-password

**Not Tested APIs**

1. GET /authn/loginAttempts/{userId} -> always return 0 attempts
2. GET /authn/token -> I do not think we can test this.
3. POST /authn/password/repeatable -> Dummy method.
4. GET /authn/log/events -> always return 0 events
5. POST /authn/log/events -> not implemented
6. DELETE /authn/log/events/{eventId} -> not implemented


mod-login-keycloak/src/main/resources/eureka/mod-login-keycloak/mod-login-keycloak-junit.feature
```diff
@@ -18,9 +18,6 @@ Feature: mod-login-keycloak integration tests setup
       | 'login.item.post'                     |
       | 'login.item.delete'                   |
       | 'login.password.validate'             |
-      | 'login.password-reset-action.post'    |
```
pkjacob, line 21: Created a dedicated user for the reset-password test, so that the 'main' user's password is not touched. So, the main user no longer need these permissions.

dmtkachenko (approved):

PR #2334, opened 2026-05-08
FAT-26033 Add member tenant item ECS query coverage

mweaver-ebsco:
This commit added C552523 coverage for member-tenant item queries against shared and local instances. Also, this commit moves central admin member-tenant capability grants into the shared consortia setup so cross-tenant query scenarios do not depend on execution order.


mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/consortia/consortia-fqm.feature
```diff
@@ -137,7 +137,7 @@ Feature: mod-consortia and mod-fqm-manager integration tests
   Scenario: Cross Tenant
     * call read('cross_tenant_et.feature')
 
-  Scenario: [Instances] [Central tenant] Query returns instances from central and member tenants
+  Scenario: Cross Tenant Instance Query
```
bvsharp: can we keep the old format here, just to keep the naming consistent with the other ECS query tests?

bvsharp (approved):

mod-fqm-manager/src/main/resources/corsair/mod-fqm-manager/features/consortia/consortia-fqm.feature
mweaver-ebsco: Sure :)

PR #2364, opened 2026-05-20
[FAT-21327] Moved another test to extended

damien-git:
## Purpose
[FAT-21327](https://folio-org.atlassian.net/browse/FAT-21327) - Thunderjet - Create Karate test - Order auto-closing when Receipt status and Payment status are set to Receipt/Payment Not Required

## Approach
Moved matching existing test to extended, with the Testrail tag.

### TODOS and Open Questions
Avoid numbering tests so we can remove them without leaving gaps or having to renumber everything.

### DIFF
--- a/acquisitions/src/test/java/org/folio/OrdersApiTest.java
+++ b/acquisitions/src/test/java/org/folio/OrdersApiTest.java
@@ -69,7 +69,6 @@ class OrdersApiTest extends TestBaseEureka implements AcquisitionsTest {
     FEATURE_43("open-order-with-manual-exchange-rate", true),
     FEATURE_44("open-order-with-many-product-ids", true),
     FEATURE_45("open-order-without-holdings", true),
-    FEATURE_46("open-order-with-resolution-statuses", true),
     FEATURE_47("open-order-with-restricted-locations", true),
     FEATURE_48("open-order-with-the-same-fund-distributions", true),
     FEATURE_49("order-event", true),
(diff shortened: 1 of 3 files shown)


acquisitions/src/test/java/org/folio/OrdersExtendedApiTest.java
```diff
@@ -231,4 +232,12 @@ void itemAppearsUnderHoldingsAfterInstanceConnectionChangeCreateNew() {
...
-}
\ No newline at end of file
+
+  @Test
+  @DisplayName("(Thunderjet) (C580268) Order auto-closing when Receipt status and Payment status are set to Receipt/Payment Not Required")
+  @EnabledIfSystemProperty(named = "test.mode", matches = "no-shared-pool")
+  void openOrderWithResolutionPoLineStatuses() {
+    runFeatureTest(Feature.FEATURE_21.getFileName());
+  }
+
```
BKadirkhodjaev: Empty line

acquisitions/src/test/java/org/folio/OrdersApiTest.java
```diff
@@ -69,7 +69,6 @@ private enum Feature implements org.folio.test.config.CommonFeature {
     FEATURE_43("open-order-with-manual-exchange-rate", true),
     FEATURE_44("open-order-with-many-product-ids", true),
     FEATURE_45("open-order-without-holdings", true),
-    FEATURE_46("open-order-with-resolution-statuses", true),
```
BKadirkhodjaev, line 72: Need to reindex now, 46 is now unallocated

damien-git: I thought you didn't want to reindex. I can do it, but eventually we should find a solution without numbers.

BKadirkhodjaev: I'm only against alphabetic reordering, it will drastically change everything, but filling in gaps is not as dangerous

BKadirkhodjaev (approved):

SerhiiNosko (approved):

PR #2384, opened 2026-05-27
Firebird: publish test results to TestRails

siarhei-charniak:
## Purpose

## Approach

### TODOS and Open Questions

### Learning


edge-oai-pmh/src/test/java/org/folio/ConsortiaEdgeOaiPmhCriticalPathApiTest.java
```diff
@@ -26,7 +27,12 @@ public void setup() {
 
     @AfterAll
     public void tearDown() {
```
SerhiiNosko, line 29: if method does not override afterAll() then super.afterAll() should be invoked automatically from TestBaseEureka, and in this case we do not need to make changes in this method, please check it to avoid duplicate execution of super.afterAll()

edge-oai-pmh/src/test/java/org/folio/ConsortiaEdgeOaiPmhCriticalPathApiTest.java
```diff
@@ -6,17 +6,18 @@
...
 import org.folio.test.config.TestModuleConfiguration;
 import org.folio.test.services.TestIntegrationService;
+import org.folio.test.services.TestRailService;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
 @FolioTest(team = "firebird", module = "edge-oai-omh")
-public class ConsortiaEdgeOaiPmhApiTest extends TestBaseEureka {
+public class ConsortiaEdgeOaiPmhCriticalPathApiTest extends TestBaseEureka {
```
SerhiiNosko, line 15: also do not forget to create PR for R1-2026 as well, because R1-2026 is used to run karate tests for Trillium

SerhiiNosko (approved):

edge-oai-pmh/src/test/java/org/folio/ConsortiaEdgeOaiPmhCriticalPathApiTest.java
SerhiiNosko: yes, only  this line should be reamined runFeature(TEST_BASE_PATH + "destroy-consortia.feature");

super.afterAll() is needed only when the current method named as afterAll()

RuslanLavrov: @siarhei-charniak, please don't forget to remove `super.afterAll();` call from the methods annotated with `@afterAll` in the test classes, if their naming is not "afterAll" (as Serhii Nosko suggested), to prevent invocation of TestBaseEureka.afterAll() two times.
This is needed for these classes as well: EdgeOaiPmhCriticalPathApiTest, ModDataExportCriticalPathApiTest, ModOaiPmhCriticalPathTests, ModOaiPmhExtendedTests

mod-data-export/src/test/java/org/folio/ModDataExportCriticalPathApiTest.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final String TEST_BASE_PATH = "classpath:firebird/dataexport/features/";
+
+    public ModDataExportCriticalPathApiTest() {
+        super(new TestIntegrationService(
+                new TestModuleConfiguration(TEST_BASE_PATH)), new TestRailService());
+    }
+
+    @BeforeAll
+    public void modDataExportTestsBeforeAll() {
+        runFeature("classpath:firebird/dataexport/data-export-basic-junit.feature");
```
RuslanLavrov, line 28: could you say please, should we set here unique values for the "testTenant" and "testTenantId" system properties, as done in the test classes of the mod-oai-pmh module, to prevent ModDataExportCriticalPathApiTest class from using deleted tenant that was created and deleted by previous test class (for example by ModDataExportApiTest class and vice versa)? Or is it handled in another way and we don't need it here?

Also, do we need to populate these properties in the EdgeOaiPmhApiTest, EdgeOaiPmhCriticalPathApiTest classes as well?

siarhei-charniak: This is impossible since test user tenants and passwords are hardcoded in ephemeral-properties file.

RuslanLavrov: could you please say what ephemeral-properties file do you mean? Is this file specified through the system properties during Karate tests run? 
Does it mean that both ModDataExportCriticalPathApiTest and ModDataExportApiTest classes will use the same testTenant from the ephemeral-properties file? I'm just asking to try to avoid probable error during regular Karate tests run after the merge

siarhei-charniak: No, it is not specified through the system properties, it is created upon environment start and resides in Storage->ConfigMaps->edge-oai-pmh-ephemeral-properties. So both ModDataExportCriticalPathApiTest and ModDataExportApiTest classes should use the same testTenant.

PR #2399, opened 2026-06-01
[FAT-21157] Recalculate budget

damien-git:
## Purpose
[FAT-21157](https://folio-org.atlassian.net/browse/FAT-21157) - Thunderjet - Create Karate test - Budget summary for Net Transfers includes Rollover transfer amounts after budget total recalculation

## Approach
Instead of implementing [C627392](https://foliotest.testrail.io/index.php?/cases/view/627392), this PR adds a new test to check recalcutating budgets, for which no test existed previously. It will cover what was meant to be tested in C627392 (that rollover transfers are taken into account when recalculating, see [MODFIN-394](https://folio-org.atlassian.net/browse/MODFIN-394)).
The test is added to core finance tests, not as an extended test.
`createTransaction` is improved to avoid using both `fromFundId` and `toFundId` with the same id.

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-finance/features/recalculate-budget.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-finance/features/recalculate-budget.feature
@@ -1,0 +1,101 @@
+# For MODFIN-394, FAT-21157
+Feature: Recalculate budget
+
+  Background:
+    * print karate.info.scenarioName
+    * url baseUrl
+
+    * callonce login testUser
+    * def okapitokenUser = okapitoken
+    * def headersUser = { 'Content-Type': 'application/json', 'x-okapi-token': '#(okapitokenUser)', 'Accept': 'application/json', 'x-okapi-tenant': '#(testTenant)' }
+    * configure headers = headersUser
+
+    * callonce variables
+
+  @Positive
+  Scenario: Recalculate budget
+    * def fundId1 = call uuid
+    * def fundId2 = call uuid
+    * def budgetId1 = call uuid
+    * def budgetId2 = call uuid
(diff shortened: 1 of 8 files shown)


acquisitions/src/main/resources/thunderjet/mod-finance/features/recalculate-budget.feature
```diff
@@ -0,0 +1,101 @@
+# For MODFIN-394, FAT-21157
```
SerhiiNosko, line 1: This test need to be declared in ExtendedFinanceApiTest

Image

Without it result will not be appearing in test run. Coupled days ago we run CriticallPath, Extended, Smoke java classes against R1-2026 and only tests.
Here is how it looks like:

Image

damien-git: I did not implement C627392. I implemented a different test which I think is more appropriate for backend tests. This is why I did not use the related annotation, and I did not include the new test in the extended tests. I would suggest not implementing C627392 at all, because the other test covers the functionality better.

SerhiiNosko: Got it, then we have 2 options. mark with this annotation some existing test or remove lable Karate automated from related test case, in any case approving

SerhiiNosko (approved):

acquisitions/src/main/resources/thunderjet/mod-finance/features/recalculate-budget.feature
damien-git: Removed "Karate Approved" from "Automation Scope".

PR #2433, opened 2026-06-10
FAT-21338 Both POL details and UUID are updated when Changing instance, items are appearing

ykashtalian:
## Purpose
[FAT-21338](https://folio-org.atlassian.net/browse/FAT-21338)  Both POL details and UUID are updated when Changing instance, items are appearing (""P/E mix"" order format)

## Approach
Implements karate tests

### TODOS and Open Questions

### Learning

### DIFF
--- a/acquisitions/src/main/resources/thunderjet/mod-orders/features/pe-mix-change-instance-connection-find-create-keep-holdings.feature
+++ b/acquisitions/src/main/resources/thunderjet/mod-orders/features/pe-mix-change-instance-connection-find-create-keep-holdings.feature
@@ -1,0 +1,210 @@
+# For MODORDERS-749, MODORDSTOR-311, MODORDSTOR-458, MODORDSTOR-481, https://foliotest.testrail.io/index.php?/cases/view/359150
+Feature: P/E Mix Change Instance Connection Find Or Create New Keep Holdings Same Location
+
+  Background:
+    * print karate.info.scenarioName
+    * url baseUrl
+
+    * callonce login testAdmin
+    * def okapitokenAdmin = okapitoken
+    * callonce login testUser
+    * def okapitokenUser = okapitoken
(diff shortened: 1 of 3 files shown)


acquisitions/src/main/resources/thunderjet/mod-orders/features/pe-mix-change-instance-connection-find-create-keep-holdings.feature
```diff
@@ -0,0 +1,210 @@
...
+    # 3.1 Wait Until POL Has Instance ID And Auto-Created Holding ID
+    * print '3.1 Wait Until POL Has Instance ID And Auto-Created Holding ID'
+    Given path 'orders/order-lines', poLineId
+    And retry until response.instanceId != null && response.locations[0].holdingId != null
+    When method GET
+    Then status 200
+    * def instanceIdOld = response.instanceId
+
+    # 3.2 Collect Both Loc1 Holdings That Exist On The Old Instance (Created During Order Open For Both Pieces)
+    * print '3.2 Collect Both Loc1 Holdings Created On The Old Instance During Order Open'
```
SerhiiNosko: we commonly do not use print for each case, only comments

SerhiiNosko (approved):

acquisitions/src/main/resources/thunderjet/mod-orders/features/pe-mix-change-instance-connection-find-create-keep-holdings.feature
damien-git: I think using both is OK, this way they show up in the logs when running JUnit tests, and also in the karate reports. The comments should match the prints though, which makes maintenance harder. I have not found a solution to use a single command for both, and at this point I think it's simply not possible.

SerhiiNosko: when we have duplicated information, it is a bit harder to read the test. and in case of failure stack trace with exact line is shown anyway, so prints is not so useful in this case

Saba-Zedginidze-EPAM (approved):

PR #2434, opened 2026-06-10
[MODORDERS-1439] Increase pause for slow execution

damien-git:
## Purpose
Increase pause for slow execution (nightly run)

## Approach
Pause 300ms instead of 100ms.
Waiting for the right status would not be a better solution, because the status changes during that time.


acquisitions/src/main/resources/thunderjet/mod-orders/helpers/helper-order-status-automatic-change.feature
SerhiiNosko, line 25: better to use retry mechanism here instead of pauses

SerhiiNosko (approved):

PR #2437, opened 2026-06-11
[MODORDERS-1456] Auto-reopen order in new FY

damien-git:
## Purpose
[MODORDERS-1456](https://folio-org.atlassian.net/browse/MODORDERS-1456) - When cancelling an invoice, order is reopened but encumbrance is not created

## Approach
- Created a new test with the 2 cases from the ticket.
- Updated delays elsewhere (they might need further updates in the future)

## Related PR
- [mod-orders](https://github.com/folio-org/mod-orders/pull/1332)


SerhiiNosko (approved):
nice test

PR #2546, opened 2026-09-09
MODUSERSKC-165 - self endpoint test

alekGbuz:
### Purpose                                                                                                                                                                                                      
  `GET /users-keycloak/_self` derives caller identity from the JWT, yet was previously declared as a public route (`permissionsRequired: []`), allowing unauthenticated callers to reach it. These tests verify that the updated wildcard descriptor (`permissionsRequired: ["*"]`) correctly rejects unauthenticated requests and continues to serve authenticated ones.

  US: [MODUSERSKC-165](https://folio-org.atlassian.net/browse/MODUSERSKC-165)

  ### Approach

  A dedicated Karate feature covers both sides of the enforcement boundary: a negative scenario that calls the endpoint without a token and expects 401, and a positive scenario that supplies a valid token and expects 200. The new feature is wired into the existing JUnit runner alongside the other mod-users-keycloak feature suites.

  - Added `self.feature` with a `@Negative` scenario that omits the token but provides a hardcoded `x-okapi-user-id` and asserts `status 401`, proving the sidecar rejects the request due to the missing token rather than a missing user ID header.
  - Added a `@Positive` scenario in `self.feature` that authenticates as `testUser`, asserts `status 200` with an active user, and cross-checks `response.user.id` against `response.permissions.userId` to confirm the sidecar resolved the correct caller identity from the JWT.
  - Added `self()` test method to `ModUsersKeycloakTests` to register the new feature with the JUnit runner.

### TODOS and Open Questions

### Learning


alekGbuz:
image

alekGbuz:
image

pkjacob (approved):

dmtkachenko (approved):

PR #2549, opened 2026-09-10
MODCONSKC-118 - self endpoint test

alekGbuz:
## Purpose

  The GET /consortia/{consortiumId}/_self endpoint in mod-consortia-keycloak is being updated to declare permissionsRequired: ["*"] in its module descriptor, requiring a valid access token from all callers. This PR adds integration test coverage to verify that unauthenticated requests are rejected with 401, and that authenticated requests continue to return the caller's consortium context.

  Jira: https://folio-org.atlassian.net/browse/MODCONSKC-118

## Approach

  A new consortia-self.feature is added with a @Negative scenario that sends a request with only x-okapi-tenant and x-okapi-user-id headers (no token) and asserts status 401, and a @Positive scenario that uses a valid consortiaAdmin token and asserts status 200 with a non-null userTenants collection in the response.
 The feature is registered in consortia-junit.feature immediately after the user-tenant-associations  block so full consortium context is established before the positive scenario runs.
### TODOS and Open Questions

### Learning

### DIFF
--- a/mod-consortia/src/main/resources/thunderjet/mod-consortia/consortia-junit.feature
+++ b/mod-consortia/src/main/resources/thunderjet/mod-consortia/consortia-junit.feature
@@ -137,11 +137,8 @@ Feature: mod-consortia-keycloak integration tests
   Scenario: User-Tenant associations api tests
     * call read('features/user-tenant-associations.feature')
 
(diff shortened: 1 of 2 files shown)


alekGbuz:
image

SerhiiNosko (approved):

PR #2559, opened 2026-09-17
[FAT-27653] Automate test cases for UXPROD-5783

ykashtalian:
## Purpose
[https://folio-org.atlassian.net/browse/FAT-27653](https://folio-org.atlassian.net/browse/FAT-27653)

## Approach
Implements karate tests
image

image

### TODOS and Open Questions

### Learning


mod-lists/src/test/java/org/folio/ModListsTest.java
```diff
@@ -79,6 +80,18 @@ void testVersioning() {
         runFeatureTest("versions");
     }
 
+    @Test
+    @DisplayName("(C1348599) Agreements + Lines ET Displays Agreement And Line Records")
```
SerhiiNosko: thinking about extracting these tests into seprate classes for Extended CriticalPath or Smoke tests. In this case they will be picked up by TestRail parser

SerhiiNosko (approved):

